PR #3, opened 2015-10-14
Add initial unittest for dynmemlib

SeppoTakalo:
This does not fully cover the dynmemlib, but it is a start.


SeppoTakalo:
@AnttiKauppila  @juhhei01 Can you review my unittests?

@mikter Can we add libService unittests to Jenkins?

AnttiKauppila:
Looking good

PR #4, opened 2015-10-14
Remove obsolete functions

SeppoTakalo:
Remove obsolete heap_scan() and heap_cor_scan() functions. They are not referred from header file.


SeppoTakalo:
@juhhei01 Can you review?

PR #5, opened 2015-10-15
Enable traces when building with yotta debug

anttiylitokola:
@jupe, can you check the change?


jupe:
this is ok for me. actually I have cloned this trace library and created seperate yotta module, so I'm not going to maintenance this anymore. new module is here: github.com/ARMmbed/mbed-client-trace , note: function names has been renamed, but tr_* macros should be backward compatible with original one.

PR #6, opened 2015-10-15
Fix trace activation.

anttiylitokola:
@SeppoTakalo, can you check the change? If it looks good can you update version and publish to yotta registry also?


SeppoTakalo:
Why on all Yotta builds?
You probably want this in mbed OS, so use #if defined(TARGET_LIKE_MBED)

anttiylitokola:
Actually we need this for linux builds as well so that's why compared against YOTTA_CFG.

SeppoTakalo:
OK, makes sense.
+1

PR #7, opened 2015-10-21
Match return values of dummy-functions.

SeppoTakalo:
This is required in order to get types match
and allow tail-calls with dummy functions.


SeppoTakalo:
@kjbracey-arm @jupe

PR #9, opened 2015-10-27
Fixed ArmCC compilation problem

AnttiKauppila:


miklis:
Automatic CI verification build not done, please verify manually.

miklis:
Automatic CI verification build not done, please verify manually.

yogpan01:
@kjbracey-arm Please review

kjbracey-arm:
Change is fine, but commit is a bit short of explanation.

If anyone had the time, it would be nice to chase up with someone why they're passing the "--gcc" option to armcc. Seems unhelpful.

0xc0170:
> Change is fine, but commit is a bit short of explanation.

Exactly, can you provide more information in the commit, what was the problem and how this one fixes it? Seems like the another paragraph explains this.

> If anyone had the time, it would be nice to chase up with someone why they're passing the "--gcc" option to armcc. Seems unhelpful.

I can't recall the exact reason. Please create an issue for gcc target, we can look at this.

kjbracey-arm:
Analysis of the compilation problem is here:

http://jira.arm.com/browse/IOTCLT-363

Basically "--gcc" makes armcc pretend really hard to be gcc - to the extent that it defines __GNUC__ and related symbols.

armcc 5.06 increased the version of gcc that it pretends to be, and that newer version claim made ns_types.h try to use some GCC extensions that armcc doesn't actually provide.

I'm not comfortable with having a non-GCC compiler set up so it defines __GNUC__. Feels like a compatibility fudge that will cause as many problems as it solves.

Not familiar enough with the Yotta build system to know which component to direct the issue to.

0xc0170:
@kjbracey-arm This is the target which defines --gnu, https://github.com/ARMmbed/target-mbed-armcc

yogpan01:
@0xc0170 Is this a problem now in target-mbed-armcc  module ?
Why is it defining --gnu for armcc ?

SeppoTakalo:
+1

yogpan01:
This fix is ok for now. We might revisit this solution again at some point of time to verify if there is any better solution for handling Static_assert for different version of compilers.

PR #12, opened 2015-11-02
Allow ns_lists of incomplete types

kjbracey-arm:
Allow declaration of a list head whose element structure type is
incomplete.

This permits better abstraction in header files - with the previous
setup all structures had to be fully visible, but you can now declare
a list of a structure that you do not have access to.

The restriction is that the link element must be the first element in
the complete structure.

Link offset type storage modified to actually permit the link element
to be the first element - it failed if the offset was 0, due to
declaring a zero-length array.

### DIFF
--- a/mbed-client-libservice/ns_list.h
+++ b/mbed-client-libservice/ns_list.h
@@ -117,49 +117,65 @@ typedef struct ns_list {
  * ~~~
  */
 #define NS_LIST_HEAD(entry_type, field) \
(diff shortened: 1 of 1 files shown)


kjbracey-arm:
@SeppoTakalo, @TeroJaasko , @juhhei01, @mikter, @jpellikk-arm

bogdanm:
+1 as far as I'm concerned :) This fixes compilation errors in C++11 mode.

PR #14, opened 2015-12-17
ns_trace: check before using the value of m_trace.line

TeroJaasko:
If the board initialization has not yet called the trace_init()
before it issued a tracef() call, the code dereferenced a null
pointer. Verify the value of m_trace.line before using it to
avoid the crash. It is easier to debug a missing tracef() than
a board crash during the initialization.

@jupe @SeppoTakalo


SeppoTakalo:
test this please

SeppoTakalo:
+1

PR #15, opened 2015-12-31
Use malloc/free in mbedOS debug build tracing (ONME-2285)

artokin:
Use malloc/free instead of ns_dyn_mem_alloc/free when tracing mbed
debug builds. nsdynmemlib is using platform_enter_critical() and
platform_exit_critical() that are not always available in mbed builds.

@SeppoTakalo , @anttiylitokola would you please review?

### DIFF
--- a/mbed-client-libservice/ns_trace.h
+++ b/mbed-client-libservice/ns_trace.h
@@ -101,12 +101,12 @@ extern "C" {
 
 #if defined  __GNUC__ || defined __CC_ARM
 /**
- * Initialize trace functionality
+ * Initialize trace functionality. This method must be called from application process.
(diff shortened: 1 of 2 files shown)


SeppoTakalo:
+0.5 but it should be documented that these functions are then not safe to be used from interrupts.

Allocating memory from interrupt handler is always a bad idea and can easily cause heap corruption.

I'm totally not sure that developers in mbed os, understand when they are in interrupt context.

artokin:
A comment added to tracing functions that are using malloc. 

The new trace library (https://github.com/ARMmbed/mbed-client-trace) seems to use malloc for all YOTTA builds.

PR #16, opened 2016-02-03
Avoid shadowed variable warning in safe foreach

kjbracey-arm:
The ns_list_foreach_safe macro used the identifier "_next" every
time, which could generate a "shadowed" variable warning if nested.

Paste the iteration pointer's name onto the _next identifier to avoid
this.


kjbracey-arm:
@AnttiKauppila, @SeppoTakalo

SeppoTakalo:
+1

PR #22, opened 2016-03-08
Fixes the memory leak and ns_dyn_mem memory corruption.

karsev:


karsev:
@SeppoTakalo

SeppoTakalo:
Is there a similar fix for master also?

PR #24, opened 2016-04-19
ONME-2549: Cope with glibc defining _Static_assert

kjbracey-arm:
Recent versions of glibc are defining _Static_assert, in a way
which breaks us. Their definition doesn't work in a struct/union,
and ns_list.h requires this.

Detect the condition, and override their definition, to get back
to the GCC 4.6 built-in keyword.

### DIFF
--- a/mbed-client-libservice/ns_types.h
+++ b/mbed-client-libservice/ns_types.h
@@ -212,6 +212,21 @@ typedef int_fast32_t int_fast24_t;
 # if __STDC_VERSION__ >= 201112L
 # define NS_STATIC_ASSERT(test, str) _Static_assert(test, str);
 # elif defined __GNUC__ && NS_GCC_VERSION >= 40600 && !defined __CC_ARM
+# ifdef _Static_assert
+    /*
+     * Some versions of glibc cdefs.h (which comes in via <stdint.h> above)
+     * attempt to define their own _Static_assert (if GCC < 4.6 or
+     * __STRICT_ANSI__) using an extern declaration, which doesn't work in a
+     * struct/union.
+     *
+     * For GCC >= 4.6 and __STRICT_ANSI__, we can do better - just use
+     * the built-in _Static_assert with __extension__. We have to do this, as
+     * ns_list.h needs to use it in a union. No way to get at it though, without
+     * overriding their define.
+     */
+#   undef _Static_assert
+#   define _Static_assert(x, y) __extension__ _Static_assert(x, y)
+# endif
 # define NS_STATIC_ASSERT(test, str) __extension__ _Static_assert(test, str);
 # else
 # define NS_STATIC_ASSERT(test, str)


kjbracey-arm:
@teetak01, @SeppoTakalo, @wipu-arm, @AnttiKauppila, @TeroJaasko

TeroJaasko:
+1. 

The NS_STATIC_ASSERT() seems to still produce quite a bit of warnings, but that is not made any worse or different with this change. It would be nice if they were fixed some point also. 

Command:  
"make -C nanostack CC=arm-none-eabi-gcc CPU=Cortex-M3"

GCC used:
"arm-none-eabi-gcc (GNU Tools for ARM Embedded Processors) 5.2.1 20151202 (release) [ARM/embedded-5-branch revision 231848]"

Output:
--8<--8<--8<--8<--8<--8<
Building source/6LoWPAN/ND/nd_router_object.c
In file included from source/nsconfig.h:12:0,
                 from source/6LoWPAN/ND/nd_router_object.c:4:
../libService/libService/ns_list.h:147:34: warning: comparison of unsigned expression >= 0 is always true [-Wtype-limits]
     NS_STATIC_ASSERT(link_offset <= UINT_FAST8_MAX, "link offset too large") \
                                  ^
../libService/libService/ns_types.h:215:67: note: in definition of macro 'NS_STATIC_ASSERT'
 # define NS_STATIC_ASSERT(test, str) __extension__ _Static_assert(test, str);
                                                                   ^
../libService/libService/ns_list.h:137:5: note: in expansion of macro 'NS_LIST_HEAD_BY_OFFSET_'
     NS_LIST_HEAD_BY_OFFSET_(entry_type, 0)
     ^
source/RPL/rpl_control.h:30:5: note: in expansion of macro 'NS_LIST_HEAD_INCOMPLETE'
     NS_LIST_HEAD_INCOMPLETE(struct rpl_instance) instances;
     ^

kjbracey-arm:
I'm stumped about how to suppress those warnings. Why would the compiler give warnings about "this condition is always true" in a compile-time assertion? Of course it's always true. Suggestions welcome.

TeroJaasko:
The warning seems to come from "0 <= UINT_FAST8_MAX" case. Compiler bug, I agree. One possible fix is below, which seems to work on my machine.

"--- ./exported-libs/mbed-client-libservice/mbed-client-libservice/ns_list.h	2016-04-19 13:10:33.033538219 +0300
+++ ./libService/ns_list.h	2016-04-19 13:03:52.057550147 +0300
@@ -144,7 +144,7 @@
 union \
 { \
     ns_list_t slist; \
-    NS_STATIC_ASSERT(link_offset <= UINT_FAST8_MAX, "link offset too large") \
+    NS_STATIC_ASSERT(link_offset <= UINT8_MAX, "link offset too large") \
     char (*offset)[link_offset + 1]; \
     entry_type *type; \
 }

teetak01:
I think those compile warnings are seperate to this PR. Could we get this part merged and could someone take the ball and make yotta-publish of new version afterwards? So far this defect is affecting at least some of our clients and with the release of stable Ubuntu 16.04 it will become more pronounced soon.

kjbracey-arm:
Agree that the warning is separate. I'll push if no further objections.

Weird warning fix. Guess it's because UINT8_MAX is just literal 255, and hence a signed int? So it's comparison of a signed integer with 0. Whereas UINT_FAST8_MAX is 0xFFFFFFFF, an unsigned int.

Guess it kind of works, but makes the warning more restrictive than it need be. Eg, pointers are bigger on ARM, so maybe on ARM offset is 300, but it's okay because uint_fast8_t is 32-bit. But on the MSP430, offset is only 240 because of smaller pointers, and uint_fast8_t is 8-bit. Warning about the offset being 300 on ARM would be pointless.

How about _Pragma("GCC diagnostic push") _Pragma("GCC diagnostic ignored \"-Wtype-limits\"") _Pragma("GCC diagnostic pop") around it?

Obviously needs to be made compiler specific - copy the example of NS_FUNNY_INTPTR_OK/RESTORE in ns_types.h, I guess.

PR #25, opened 2016-04-20
Suppress type-limits warnings from GCC

kjbracey-arm:
GCC produces a silly "comparison is always true" if an ns_list
has its link at offset 0, and UINT_FAST8_MAX is 0xFFFFFFFF,
and hence unsigned.

Suppress the warning.


kjbracey-arm:
@teetak01, @SeppoTakalo, @wipu-arm, @AnttiKauppila, @TeroJaasko

TeroJaasko:
Those pesky warnings are now gone. +1.

PR #28, opened 2016-04-29
Added mbed-trace submodule and adapted Makefile.

tommikas:
This is to allow compilation of applications in nanomesh-applications that depend on mbed-trace.

Next step would be to map calls made to old libTrace functions to the corresponding mbed-trace functions, and to adapt yotta build for this module to also build the submodule.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "source/mbed-trace"]
+	path = source/mbed-trace
+	url = git@github.com:ARMmbed/mbed-trace.git
--- a/.yotta_ignore
+++ b/.yotta_ignore
@@ -1,1 +1,2 @@
 test/*
+source/mbed-trace
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,14 @@ source/libip6string/stoip6.c \
 source/libList/ns_list.c \
 source/libTrace/ns_trace.c \
 source/nsdynmemLIB/nsdynmemLIB.c \
+source/mbed-trace/source/mbed_trace.c \
+source/mbed-trace/source/mbed_trace_ip6tos.c \
 
 LIB := libservice.a
-EXPORT_HEADERS := mbed-client-libservice
 
-include ../exported_rules.mk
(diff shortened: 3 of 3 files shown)


tommikas:
@jupe 

@yogpan01 This version combined either with ARMmbed/mbed-trace#20, or adding -DYOTTA_CFG_MBED_TRACE for example in Projects/makefiles/linux/simulator/Makefile should allow compiling mbed-trace dependent applications in nanomesh-applications.

yogpan01:
lgtm

jupe:
lgtm

tommikas:
@SeppoTakalo You should get the last word on this.

jupe:
package.json should be updated as well to use mbed-trace -module as dependency

SeppoTakalo:
+1

SeppoTakalo:
@jupe Do you mean `module.json`? If so, then yes, I agree, mbed-trace should be also added there as a dependency. (Or is this directly depending?)

jupe:
actually because libService doesn't use directly mbed-trace -library so perhaps it shouldn't be here (like it doesn't right now) .

tommikas:
That makes sense.

The point of this was to make mbed-trace library available to applications compiled in nanomesh-applications. To achieve that, mbed-trace could be included as a submodule in ARMmbed/nanomesh-libservice-root-private/exported-libs, i.e. at the same level in the file hierarchy as nanostack-libservice itself. Would that make sense?

SeppoTakalo:
Maybe we can add it into the libService-root as a submodule. to same level as other exported libraries are. it is not a Yotta module as itself, just a container that pulls in other exported libraries for our Makefile based builds.

See: https://github.com/ARMmbed/nanomesh-libservice-root-private

tommikas:
If I understood this correctly, that is exactly what I meant as well (but failed to link the repository).

tommikas:
Done with ARMmbed/nanomesh-libservice-root-private#27 instead.

PR #30, opened 2016-05-15
IAR support - remove array size

c1728p9:
Remove the array size parameter since this is not valid syntax for IAR.

### DIFF
--- a/mbed-client-libservice/ip_fsc.h
+++ b/mbed-client-libservice/ip_fsc.h
@@ -22,8 +22,8 @@
 #define NEXT_HEADER_UDP     0x11
 #define NEXT_HEADER_ICMP6   0x3A
 
-extern uint16_t ip_fcf_v(uint_fast8_t count, const ns_iovec_t vec[static count]);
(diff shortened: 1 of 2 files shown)


mbed-client-libservice/ip_fsc.h
```diff
@@ -22,8 +22,8 @@
 #define NEXT_HEADER_UDP     0x11
 #define NEXT_HEADER_ICMP6   0x3A
 
-extern uint16_t ip_fcf_v(uint_fast8_t count, const ns_iovec_t vec[static count]);
```
c1728p9: @kjbracey-arm, is the dynamic array size passed in used to help catch out of bounds errors?  If so, would it be preferable to change this to a macro (which can be defined away for IAR)?  Something like this:
```
extern uint16_t ip_fcf_v(uint_fast8_t count, const ns_iovec_t vec[array_size(count)]);
```

Let me know, what you think and if so I'll update the PR.

c1728p9:
This patch is no longer needed since this feature can be enabled for IAR:
https://github.com/ARMmbed/mbed-os/pull/114

PR #31, opened 2016-05-16
Minor updates to NVM API

artokin:
-Use uint32_t instead of int when dealing with data length.
-Remove KEY_NOT_FOUND from read/write api (used via callback).

### DIFF
--- a/mbed-client-libservice/platform/arm_hal_nvm.h
+++ b/mbed-client-libservice/platform/arm_hal_nvm.h
@@ -84,7 +84,7 @@ platform_nvm_status platform_nvm_finalize(nvm_callback *callback, void *context)
  * \return NVM_ERROR if key creation failed and callback will not be called.
  * \return Provided callback function will be called once operation is completed.
  */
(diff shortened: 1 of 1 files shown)


artokin:
@terhei , @mikter would you please review?

kjbracey-arm:
We tend to shy away from uint32_t on size APIs - it's unnecessarily onerous on 8/16-bit platforms.

If you are happy being limited to 64K, I'd say uint16_t, like so much stuff does (eg buffer_t, socket APIs).

If you really do want large sizes, use size_t. That at least gives the option of being 16-bit on 16-bit platforms. (But we've tended to avoid it because sometimes its 32-bit on a platform with 16-bit CPU, and we often know we never need more than 64k).

kjbracey-arm:
While you're touching it (missed it on first review) - nvm_write presumably wants a const data pointer.

And if you're going to standard size_t, you could maybe go for standard void * data pointers to match? Our conventional use of "uint8_t *" is a bit odd - may be convenient for the function implementer, less so for the users who may be forced to cast. A quick conversion at the top of the implementation covers it up:

    void my_public_api(const void *data_param) 
    {
        const uint8_t *data = data_param;

PR #33, opened 2016-05-23
Clean up & fix the alloc_sector count when using a slightly bigger bl…

SeppoTakalo:
…ock.

@AnttiKauppila

### DIFF
--- a/source/nsdynmemLIB/nsdynmemLIB.c
+++ b/source/nsdynmemLIB/nsdynmemLIB.c
@@ -203,12 +203,10 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
                     retval = 0;
                     break;
                 } else if (s_size < 0) {
+                    s_size = -s_size;
                     //Free
-                    if (((-s_size) >= alloc_sector)) {
+                    if (s_size >= alloc_sector) {
                         /*found block*/
-                        //Convert Current size
-                        s_size = -s_size;
-
                         if (s_size > (alloc_sector + 4)) {
                             //debug("Bigger\r\n");
                             //Set sector start len & Move pointer to Allocated data sector
@@ -223,30 +221,23 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
                             //Now set new slice start & End len fields
                             //Move pointer to New Free Sector
                             //Calculate new sector size
-                            alloc_sector += 2; //
-
-                            s_size = (s_size - alloc_sector);
+                            s_size = (s_size - alloc_sector - 2);
 
                             *ptr-- = -(s_size);
                             ptr -= s_size;
                             *ptr = -(s_size);
-                            if (mem_stat_info_ptr) {
-                                alloc_sector -= 2;
-                            }
                         } else {
                             //debug("USE Same\r\n");
                             //Set sector start len & Move pointer to Allocated data sector
                             *ptr = s_size;
                             ptr -= s_size;
+                            alloc_sector = s_size;
                             retval = (void *) ptr;
                             ptr--;
                             //Set End length
                             *ptr-- = s_size;
                         }
                         break;
-                    } else {
-                        //Set Number from negative to Positive
-                        s_size = -s_size;
                     }
                 }
                 //Move sector to next
@@ -316,11 +307,11 @@ void *ns_dyn_mem_temporary_alloc(int16_t alloc_size)
                     retval = 0;
                     break;
                 } else if (s_size < 0) {
+                    s_size = -s_size;
                     //Free
-                    if (((-s_size) >= alloc_sector)) {
+                    if (s_size >= alloc_sector) {
                         /*found block*/
-                        //Convert Current size
-                        s_size = -s_size;
+
                         if (s_size > (alloc_sector + 4)) {
                             //debug("Bigger\r\n");
                             //Set sectror start len & Move pointer to Allocated data sector
@@ -332,15 +323,11 @@ void *ns_dyn_mem_temporary_alloc(int16_t alloc_size)
                             //Now set new slice start & End len fields
                             //Move pointer to New Free Sector
                             //Calculate new sector size
-                            alloc_sector += 2; //
-                            s_size = (s_size - alloc_sector);
+                            s_size = (s_size - alloc_sector -2);
 
                             *ptr++ = -(s_size);
                             ptr += s_size;
                             *ptr = -(s_size);
-                            if (mem_stat_info_ptr) {
-                                alloc_sector -= 2;
-                            }
                         } else {
                             //debug("USE Same\r\n");
                             //Move pointer to Allocated data sector
@@ -349,13 +336,10 @@ void *ns_dyn_mem_temporary_alloc(int16_t alloc_size)
                             retval = (void *) ptr;
                             ptr += s_size;//Move Pointer  to next free place
                             *ptr = s_size;
+                            alloc_sector = s_size;
                         }
                         break;
-                    } else {
-                        //Set Number from negative to Positive
-                        s_size = -s_size;
                     }
-
                 }
                 //Allocated sector
                 //Move
--- a/source/nsdynmemLIB/nsdynmemLIB.c
+++ b/source/nsdynmemLIB/nsdynmemLIB.c
@@ -133,14 +133,15 @@ static int heap_alloc_internal_check(int16_t alloc_size)
 
     if (alloc_size) {
         //
-        temp_sector = alloc_size;
-        temp_int = (temp_sector % temp_size);
-        if (temp_int) {
-            temp_sector += (temp_size - temp_int);
-        }
+        temp_sector = (alloc_size+temp_size-1)/temp_size;
+//        temp_sector = alloc_size;
+//        temp_int = (temp_sector % temp_size);
+//        if (temp_int) {
+//            temp_sector += (temp_size - temp_int);
+//        }
 
 
-        temp_sector /= temp_size;
+//        temp_sector /= temp_size;
 
         temp_int = (heap_size / temp_size);
         temp_int -= 2;
@@ -155,40 +156,48 @@ static int heap_alloc_internal_check(int16_t alloc_size)
 static int8_t ns_sector_validate(int *sector_start, int direction)
 {
     int8_t ret_val = -1;
-    int size_start, size_end;
-    if (direction) {
-        //Up
-        size_start = *sector_start++;
-        if (size_start < 0) {
-            sector_start += -(size_start);
-        } else {
-            sector_start += size_start;
-        }
-        size_end = *sector_start;
-    } else {
-        size_end = *sector_start--;
-        if (size_end < 0) {
-            sector_start -= -(size_end);
-        } else {
-            sector_start -= size_end;
-        }
-        size_start = *sector_start;
-    }
-    if (size_start == size_end) {
+    int size_start;
+//    int size_end;
+    int multiplier = direction?1:-1;
+    size_start = *sector_start;
+    sector_start += multiplier;
+    multiplier = size_start<0?-multiplier:multiplier;
+    sector_start += multiplier*size_start;
+
+//    if (direction) {
+//        //Up
+//        size_start = *sector_start++;
+//        if (size_start < 0) {
+//            sector_start += -(size_start);
+//        } else {
+//            sector_start += size_start;
+//        }
+//        size_end = *sector_start;
+//    } else {
+//        size_end = *sector_start--;
+//        if (size_end < 0) {
+//            sector_start -= -(size_end);
+//        } else {
+//            sector_start -= size_end;
+//        }
+//        size_start = *sector_start;
+//    }
+//    if (size_start == size_end) {
+    if (size_start == *sector_start) {
         ret_val = 0;
     }
     return ret_val;
 }
 #endif
 
-void *ns_dyn_mem_alloc(int16_t alloc_size)
+static void *ns_dyn_mem_internal_alloc(int16_t alloc_size, bool dir_up)
 {
-#ifndef STANDARD_MALLOC
-
-    int *ptr = heap_main_end;
+    int *ptr = dir_up?heap_main:heap_main_end;
+    int *end = dir_up?heap_main_end:heap_main;
     void *retval = 0;
     int h_size, alloc_sector, s_size;
     int moved = 0;
+    int multiplier = dir_up?1:-1;
 
     alloc_sector = heap_alloc_internal_check(alloc_size);
     if (alloc_sector) {
@@ -196,8 +205,9 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
         platform_enter_critical();
 
         while (moved < h_size) {
-            if (ns_sector_validate(ptr, 0) == 0) {
+            if (ns_sector_validate(ptr, dir_up) == 0) {
                 s_size = *ptr;
+
                 if (s_size == 0) {
                     heap_failure(NS_DYN_MEM_HEAP_SECTOR_CORRUPTED);
                     retval = 0;
@@ -207,35 +217,51 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
                     //Free
                     if (s_size >= alloc_sector) {
                         /*found block*/
+
                         if (s_size > (alloc_sector + 4)) {
                             //debug("Bigger\r\n");
                             //Set sector start len & Move pointer to Allocated data sector
                             *ptr = alloc_sector;
-                            //Move Pointer  to next free place
-                            ptr -= alloc_sector;
-                            //Set return Pointer
-                            retval = (void *) ptr;
-                            ptr--;
-                            //Set End length
-                            *ptr-- = alloc_sector;
+                            if (dir_up) {
+                                ptr++;
+                                //Set return Pointer
+                                retval = (void *) ptr;
+                                ptr += alloc_sector;
+                                //Set End length
+                                *ptr++ = alloc_sector;
+                            } else {
+                                ptr -= alloc_sector;
+                                //Set return Pointer
+                                retval = (void *) ptr;
+                                ptr--;
+                                //Set End length
+                                *ptr-- = alloc_sector;
+                            }
+
                             //Now set new slice start & End len fields
                             //Move pointer to New Free Sector
                             //Calculate new sector size
                             s_size = (s_size - alloc_sector - 2);
 
-                            *ptr-- = -(s_size);
-                            ptr -= s_size;
+                            *ptr = -(s_size);
+                            ptr += multiplier*(s_size+1);
                             *ptr = -(s_size);
                         } else {
                             //debug("USE Same\r\n");
                             //Set sector start len & Move pointer to Allocated data sector
                             *ptr = s_size;
-                            ptr -= s_size;
-                            alloc_sector = s_size;
-                            retval = (void *) ptr;
-                            ptr--;
+                            if (dir_up) {
+                                ptr++;
+                                retval = (void *) ptr;
+                                ptr += s_size;
+                            } else {
+                                ptr -= s_size;
+                                retval = (void *) ptr;
+                                ptr--;
+                            }
                             //Set End length
-                            *ptr-- = s_size;
+                            *ptr = s_size;
+                            alloc_sector = s_size;
                         }
                         break;
                     }
@@ -243,9 +269,8 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
                 //Move sector to next
                 //Allocated sector
                 moved += (s_size + 2);
-                ptr -= 2; //Skip Both of length fileds
-                ptr -= s_size;
-                if (heap_main == ptr) {
+                ptr += multiplier*(s_size+2);
+                if (end == ptr) {
                     break;
                 }
             } else {
@@ -257,7 +282,7 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
 
 
         if (mem_stat_info_ptr) {
-            alloc_size = (alloc_sector * sizeof(int));
+            alloc_size = ((alloc_sector+2) * sizeof(int));
             if (retval) {
                 //Update Allocate OK
                 dev_stat_update(DEV_HEAP_ALLOC_OK, alloc_size);
@@ -270,6 +295,97 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
         platform_exit_critical();
     }
     return retval;
+}
+
+void *ns_dyn_mem_alloc(int16_t alloc_size)
+{
+#ifndef STANDARD_MALLOC
+    return ns_dyn_mem_internal_alloc(alloc_size, false);
+//    int *ptr = heap_main_end;
+//    void *retval = 0;
+//    int h_size, alloc_sector, s_size;
+//    int moved = 0;
+
+//    alloc_sector = heap_alloc_internal_check(alloc_size);
+//    if (alloc_sector) {
+//        h_size = (heap_size / sizeof(int));
+//        platform_enter_critical();
+
+//        while (moved < h_size) {
+//            if (ns_sector_validate(ptr, 0) == 0) {
+//                s_size = *ptr;
+//                if (s_size == 0) {
+//                    heap_failure(NS_DYN_MEM_HEAP_SECTOR_CORRUPTED);
+//                    retval = 0;
+//                    break;
+//                } else if (s_size < 0) {
+//                    s_size = -s_size;
+//                    //Free
+//                    if (s_size >= alloc_sector) {
+//                        /*found block*/
+//                        if (s_size > (alloc_sector + 4)) {
+//                            //debug("Bigger\r\n");
+//                            //Set sector start len & Move pointer to Allocated data sector
+//                            *ptr = alloc_sector;
+//                            //Move Pointer  to next free place
+//                            ptr -= alloc_sector;
+//                            //Set return Pointer
+//                            retval = (void *) ptr;
+//                            ptr--;
+//                            //Set End length
+//                            *ptr-- = alloc_sector;
+//                            //Now set new slice start & End len fields
+//                            //Move pointer to New Free Sector
+//                            //Calculate new sector size
+//                            s_size = (s_size - alloc_sector - 2);
+
+//                            *ptr-- = -(s_size);
+//                            ptr -= s_size;
+//                            *ptr = -(s_size);
+//                        } else {
+//                            //debug("USE Same\r\n");
+//                            //Set sector start len & Move pointer to Allocated data sector
+//                            *ptr = s_size;
+//                            ptr -= s_size;
+//                            alloc_sector = s_size;
+//                            retval = (void *) ptr;
+//                            ptr--;
+//                            //Set End length
+//                            *ptr-- = s_size;
+//                        }
+//                        break;
+//                    }
+//                }
+//                //Move sector to next
+//                //Allocated sector
+//                moved += (s_size + 2);
+//                ptr -= 2; //Skip Both of length fileds
+//                ptr -= s_size;
+//                if (heap_main == ptr) {
+//                    break;
+//                }
+//            } else {
+//                heap_failure(NS_DYN_MEM_HEAP_SECTOR_CORRUPTED);
+//                retval = 0;
+//                break;
+//            }
+//        }
+
+
+//        if (mem_stat_info_ptr) {
+//            alloc_size = ((alloc_sector+2) * sizeof(int));
+//            if (retval) {
+//                //Update Allocate OK
+//                dev_stat_update(DEV_HEAP_ALLOC_OK, alloc_size);
+
+//            } else {
+//                //Update Allocate Fail
+//                dev_stat_update(DEV_HEAP_ALLOC_FAIL, alloc_size);
+//            }
+//        }
+//        platform_exit_critical();
+//    }
+//    return retval;
 #else
     void *retval = 0;
     if (alloc_size) {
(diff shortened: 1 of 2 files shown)


AnttiKauppila:
Re visited some code and refactored couple of things.
Commented out code was left intentionally if there are some findings.

@SeppoTakalo @juhhei01 Please review this.

AnttiKauppila:
smoketests passed

juhhei01:
What is test coverage?

AnttiKauppila:
96.4% I suspect that it is dead code, but we can check that later

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -133,14 +133,15 @@ static int heap_alloc_internal_check(int16_t alloc_size)
 
     if (alloc_size) {
         //
-        temp_sector = alloc_size;
-        temp_int = (temp_sector % temp_size);
-        if (temp_int) {
-            temp_sector += (temp_size - temp_int);
-        }
+        temp_sector = (alloc_size+temp_size-1)/temp_size;
```
kjbracey-arm: Maths looks correct.
Lose the "temp_size" - just use "sizeof(int)" - it will be clearer.
Remove the commented code.
Get the formatting right - spaces around operators.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -133,14 +133,15 @@ static int heap_alloc_internal_check(int16_t alloc_size)
...
+//        temp_int = (temp_sector % temp_size);
+//        if (temp_int) {
+//            temp_sector += (temp_size - temp_int);
+//        }
 
 
-        temp_sector /= temp_size;
+//        temp_sector /= temp_size;
 
         temp_int = (heap_size / temp_size);
```
kjbracey-arm: Can you tidy this too? It's basically "if size > (heap_size - 2 words), not valid". Could do that at the top of the function in bytes next to the other size check, rather than in words.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -155,49 +156,58 @@ static int heap_alloc_internal_check(int16_t alloc_size)
...
-            sector_start -= -(size_end);
-        } else {
-            sector_start -= size_end;
-        }
-        size_start = *sector_start;
-    }
-    if (size_start == size_end) {
+    int size_start;
+//    int size_end;
+    int multiplier = direction?1:-1;
```
kjbracey-arm: As it's an internal function, why not make the callers pass -1 or +1, save the extra work.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -155,49 +156,58 @@ static int heap_alloc_internal_check(int16_t alloc_size)
...
+    size_start = *sector_start;
+    sector_start += multiplier;
+    multiplier = size_start<0?-multiplier:multiplier;
```
kjbracey-arm: Overly complicated, and badly formatted. Flipping the "direction" variable based on the free status is weird.

You're just wanting to get the size of the block (independent of free status), which is abs(size_start).
I suggest (assuming direction modified to be +1 or -1):

    size = abs(*sector_start);
    sector_start += (1 + size + 1) * direction;

Which I think is fairly self-documenting. (Could someone maybe add a comment somewhere actually explaining the heap format?)

That does assume multiply-by-variable is fast though. I guess it should be on anything except poxy M0 devices. If we cared about slow-multiply devices, then it could be:

    size = abs(*sector_start);
    if (direction > 0) {
            sector_start += 1 + size + 1;
    } else {
            sector_start -= 1 + size + 1;
    }

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -155,49 +156,58 @@ static int heap_alloc_internal_check(int16_t alloc_size)
...
 }
 #endif
 
-void *ns_dyn_mem_alloc(int16_t alloc_size)
+static void *ns_dyn_mem_internal_alloc(int16_t alloc_size, bool dir_up)
 {
-#ifndef STANDARD_MALLOC
-
-    int *ptr = heap_main_end;
+    int *ptr = dir_up?heap_main:heap_main_end;
```
kjbracey-arm: Like the concept, but suggest passing +1/-1 in the first place, if you're going to do the multiplication thing.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -207,45 +217,60 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
...
                             //Set sector start len & Move pointer to Allocated data sector
                             *ptr = alloc_sector;
-                            //Move Pointer  to next free place
-                            ptr -= alloc_sector;
-                            //Set return Pointer
-                            retval = (void *) ptr;
-                            ptr--;
-                            //Set End length
-                            *ptr-- = alloc_sector;
+                            if (dir_up) {
```
kjbracey-arm: Seems a bit laboured, especially compared to the "hole"-creating bit below. 

Can't you actually unify the two cases anyway - at the minute it's 

    if (making hole) {
          make block
          make hole
    } else {
          make block
    }

Would make sense to avoid repeating the block creation. Can't quite write it out, but something like

   if (space for hole) {
         create hole (now at front, rather than back, because it's easier)
    }
   else {
        increase alloc_sector to take extra space
   }
   create block

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -207,45 +217,60 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
...
+
                             //Now set new slice start & End len fields
                             //Move pointer to New Free Sector
                             //Calculate new sector size
                             s_size = (s_size - alloc_sector - 2);
 
-                            *ptr-- = -(s_size);
-                            ptr -= s_size;
+                            *ptr = -(s_size);
+                            ptr += multiplier*(s_size+1);
```
kjbracey-arm: Where did the +1 come from here?

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -257,7 +282,7 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
 
 
         if (mem_stat_info_ptr) {
-            alloc_size = (alloc_sector * sizeof(int));
+            alloc_size = ((alloc_sector+2) * sizeof(int));
```
kjbracey-arm: Don't think this works - alloc_sector reaches here as 0 if the allocation fails (I think) so this would increase allocation on failure.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -270,6 +295,97 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
         platform_exit_critical();
     }
     return retval;
+}
+
+void *ns_dyn_mem_alloc(int16_t alloc_size)
+{
+#ifndef STANDARD_MALLOC
+    return ns_dyn_mem_internal_alloc(alloc_size, false);
+//    int *ptr = heap_main_end;
```
kjbracey-arm: Don't leave the old code.

AnttiKauppila:
I have gone through Kevin's comments and also refactored the code even more to be more simple.
@SeppoTakalo @kjbracey-arm @juhhei01 Please re-review this

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -117,77 +115,51 @@ void dev_stat_update(mem_stat_update_t type, int16_t size)
     }
 }
 
-//Fix return value allways good fo aligment
+//Fix return value to be always good for alignment
```
kjbracey-arm: Bit misleading - it returns number of sectors, rather than an aligned byte size. Although that's far more obvious from the clearer code anyway.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -117,77 +115,51 @@ void dev_stat_update(mem_stat_update_t type, int16_t size)
...
 static int heap_alloc_internal_check(int16_t alloc_size)
 {
-    int temp_int, temp_sector = 0, temp_size;
-    temp_size = sizeof(int);
+    int temp_sector = 0;
```
kjbracey-arm: Get rid of this silly "name all local variables as temp-something" convention.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -117,77 +115,51 @@ void dev_stat_update(mem_stat_update_t type, int16_t size)
...
+    } else if (alloc_size > (heap_size - 2 * sizeof(int)) ) {
+        heap_failure(NS_DYN_MEM_ALLOCATE_SIZE_NOT_VALID);
+    } else if (alloc_size) {
+        temp_sector = (alloc_size + sizeof(int) - 1) / sizeof(int);
     }
     return temp_sector;
 }
 
+// Checks that sector length indicators equals
+// Sector has format: Length (word) | data area in words | Length (word)
```
kjbracey-arm: What does "sector" actually mean in this code? I was reading it as "word", up to this comment, thinking "alloc_sector" meant "allocation size in words", but I guess it's actually "block"? Can we change it to "block"? A "sector" is usually fixed-size - it's misleading.

Other notes on the format: both lengths are negated if free, and the stored length is the size of the data area in words, not the size of the block.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -117,77 +115,51 @@ void dev_stat_update(mem_stat_update_t type, int16_t size)
...
-        size_start = *sector_start++;
-        if (size_start < 0) {
-            sector_start += -(size_start);
-        } else {
-            sector_start += size_start;
-        }
-        size_end = *sector_start;
+    int size_start = *sector_start;
+    if (direction > 0) {
+        sector_start += (1 + abs(size_start));
```
kjbracey-arm: Incrementing sector_start is weird. Because it won't point to the start any more. Make a new variable.

    sector_end = sector_start + (1 + abs(size_start));

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -117,77 +115,51 @@ void dev_stat_update(mem_stat_update_t type, int16_t size)
...
-            sector_start -= -(size_end);
-        } else {
-            sector_start -= size_end;
-        }
-        size_start = *sector_start;
+        sector_start -= (1 + abs(size_start));
     }
-    if (size_start == size_end) {
+
+    if (size_start && size_start == *sector_start) {
```
kjbracey-arm: Why "if size_start"?

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -117,77 +115,51 @@ void dev_stat_update(mem_stat_update_t type, int16_t size)
...
 #endif
 
-void *ns_dyn_mem_alloc(int16_t alloc_size)
+// For direction, use 1 for direction up and -1 for down
+static void *ns_dyn_mem_internal_alloc(int16_t alloc_size, int direction)
 {
-#ifndef STANDARD_MALLOC
-
-    int *ptr = heap_main_end;
+    int *ptr = direction > 0?heap_main:heap_main_end;
```
kjbracey-arm: Formatting.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -117,77 +115,51 @@ void dev_stat_update(mem_stat_update_t type, int16_t size)
...
+    int *end = direction > 0?heap_main_end:heap_main;
+    int *temp_ptr;
```
kjbracey-arm: Don't declare this outside the block it's used in.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -196,67 +168,43 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
...
+                        /*found block*/
                         if (s_size > (alloc_sector + 4)) {
-                            //debug("Bigger\r\n");
-                            //Set sector start len & Move pointer to Allocated data sector
-                            *ptr = alloc_sector;
-                            //Move Pointer  to next free place
-                            ptr -= alloc_sector;
-                            //Set return Pointer
+                            //There is enough room for new hole so create it first
+                            temp_ptr = (ptr + ((alloc_sector + 2) * direction));
```
kjbracey-arm: "hole_ptr" would be a better name. Or maybe "free_ptr". Make them match, anyway.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -196,67 +168,43 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
...
+                            free_size = s_size - alloc_sector - 2;
+                            *temp_ptr = -(free_size);
+                            temp_ptr += direction * (free_size + 1);
```
kjbracey-arm: Be consistent about multiplying with direction or not. Do it everywhere or nowhere.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -196,67 +168,43 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
...
-                            //debug("Bigger\r\n");
-                            //Set sector start len & Move pointer to Allocated data sector
-                            *ptr = alloc_sector;
-                            //Move Pointer  to next free place
-                            ptr -= alloc_sector;
-                            //Set return Pointer
+                            //There is enough room for new hole so create it first
+                            temp_ptr = (ptr + ((alloc_sector + 2) * direction));
+                            free_size = s_size - alloc_sector - 2;
+                            *temp_ptr = -(free_size);
```
kjbracey-arm: Pointless brackets

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -196,67 +168,43 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
...
-
-                            s_size = (s_size - alloc_sector);
-
-                            *ptr-- = -(s_size);
-                            ptr -= s_size;
-                            *ptr = -(s_size);
-                            if (mem_stat_info_ptr) {
-                                alloc_sector -= 2;
-                            }
+                            ptr += alloc_sector;
```
kjbracey-arm: Don't need the (void *) cast.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -196,67 +168,43 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
...
                 }
                 //Move sector to next
                 //Allocated sector
                 moved += (s_size + 2);
-                ptr -= 2; //Skip Both of length fileds
-                ptr -= s_size;
-                if (heap_main == ptr) {
-                    break;
-                }
+                ptr += direction*(s_size + 2);
```
kjbracey-arm: Formatting

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -266,19 +214,28 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
 
 
         if (mem_stat_info_ptr) {
-            alloc_size = (alloc_sector * sizeof(int));
+            // If alloc fail needs to collect size of failed alloc,
+            // this code needs to be revisited
+            alloc_size = ((alloc_sector + 2) * sizeof(int));
```
kjbracey-arm: Interesting. How do we want failures logged? At present each failure adds "rounded up size". Adding 0 doesn't work - it doesn't affect the stat!

You could just update by 1, but I guess I'd try to roughly retain previous behaviour by returning the /original/ alloc request size.

So move this into the if (ret_val), and let the failure return the /original/ alloc_size passed in.
Better still, don't actually modify the original alloc_size (it's confusing) - just put the expression in the OK stat update.

Maybe worth marking the alloc_size parameter "const" to discourage internal tinkering.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -266,19 +214,28 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
...
             }
         }
         platform_exit_critical();
     }
     return retval;
+}
+
+void *ns_dyn_mem_alloc(int16_t alloc_size)
+{
+#ifndef STANDARD_MALLOC
```
kjbracey-arm: Could you not have a single STANDARD_MALLOC ifdef inside the internal function? Save having two copies of the "call malloc" code?

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -401,67 +270,38 @@ void *ns_dyn_mem_temporary_alloc(int16_t alloc_size)
...
-
-            cur_len = *cur_sector;
-            cur_len = -cur_len; //Invert
-            cur_sector += (cur_len + 1); //New End pointer
-            prev_len += (cur_len + 2); // SET NEW LEN
-        } else {
-            prev_sector = 0;
+        cur_sector--;
+        if (*cur_sector < 0) {
+            total_size += (2 - *cur_sector);
```
kjbracey-arm: Half-inclined to write "2 + (-*cur_sector)" to make clearer the negative size thing, but I guess it's fairly obvious.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -401,67 +270,38 @@ void *ns_dyn_mem_temporary_alloc(int16_t alloc_size)
...
+    if (end != heap_main_end) {
+        end++;
+        if (*end < 0) {
+            total_size += (2 - *end);
+            end += (1 - *end);
+        }else{
+            end--;
+        }
     }
+    *cur_sector = -total_size;
```
kjbracey-arm: Why write to *cur_sector? Either it's the same as start, in which case it's pointless, or it's in the middle of a larger reason - still pointless.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -493,24 +333,17 @@ void ns_dyn_mem_free(void *block)
...
-            //Check  prev sector
-            //Try Link Next between cur
-            ns_free_and_merge_next_sector(ptr, size);
-            //Try Merge Prev with cur
-            ns_merge_prev_sector(ptr);
+            ns_free_and_merge_with_adjacent_sectors(ptr, size);
             if (mem_stat_info_ptr) {
                 //Update Free Counter
-                size = (size * sizeof(int));
+                size = ((size+2) * sizeof(int));
```
kjbracey-arm: Formatting. And excess brackets. As previous, I would prefer just using an expression to actually changing the variable.

Very hard to follow if one variable changes units or fundamental meaning throughout a function.

kjbracey-arm:
Looking a lot better now. Nearly there.

source/nsdynmemLIB/nsdynmemLIB.c
AnttiKauppila: Size must be non-zero

source/nsdynmemLIB/nsdynmemLIB.c
AnttiKauppila: I rename free_size to hole_size. It is current block size - used size => no multiplication

source/nsdynmemLIB/nsdynmemLIB.c
AnttiKauppila: This code was there earlier also, it prevents double deletion of the pointer.

source/nsdynmemLIB/nsdynmemLIB.c
kjbracey-arm: Ah, got it. Was confused because the test on size was done after using the size.

I'd be inclined to move the size == 0 rejection to the top. Particularly as you probably should also be more strict on the test there before trying to follow advance the pointer.

Two other corruption indications are abs(size_start) being > (heap_size / sizeof(int) -  2),  or size being INT_MIN, which would stop abs() working - check that first.

And when doing numeric comparisons, don't use boolean shorthands. It's clearer to write the comparison with zero, rather than taking advantage of zero having the special meaning "false" - most particularly don't mix implicit boolean tests with comparisons, as this "if" does.

source/nsdynmemLIB/nsdynmemLIB.c
kjbracey-arm: That's a good reason. Add an explanatory comment.

AnttiKauppila:
@kjbracey-arm Please revisit
smoketests are passing locally

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -117,146 +114,96 @@ void dev_stat_update(mem_stat_update_t type, int16_t size)
     }
 }
 
-//Fix return value allways good fo aligment
 static int heap_alloc_internal_check(int16_t alloc_size)
 {
-    int temp_int, temp_sector = 0, temp_size;
-    temp_size = sizeof(int);
+    int block_count = 0;
```
kjbracey-arm: Can we rename this function. It isn't really a "check", apart from as a side effect. It mainly converts requested size to size in words.

     static int convert_allocation_size(int16_t requested_bytes)

or something like that?

And make the return variable "words" or something - block count sounds to me like "number of blocks".

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -117,146 +114,96 @@ void dev_stat_update(mem_stat_update_t type, int16_t size)
...
-
-        temp_int = (heap_size / temp_size);
-        temp_int -= 2;
-        if (temp_sector > temp_int) {
-            heap_failure(NS_DYN_MEM_ALLOCATE_SIZE_NOT_VALID);
-            temp_sector = 0;
-        }
+    } else if (alloc_size > (heap_size - 2 * sizeof(int)) ) {
+        heap_failure(NS_DYN_MEM_ALLOCATE_SIZE_NOT_VALID);
+    } else if (alloc_size) {
```
kjbracey-arm: Unneeded if.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -117,146 +114,96 @@ void dev_stat_update(mem_stat_update_t type, int16_t size)
...
+        block_count = (alloc_size + sizeof(int) - 1) / sizeof(int);
     }
-    return temp_sector;
+    return block_count;
 }
 
-static int8_t ns_sector_validate(int *sector_start, int direction)
+// Checks that block length indicators equals
```
kjbracey-arm: "make sense" rather than "equals"?

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -117,146 +114,96 @@ void dev_stat_update(mem_stat_update_t type, int16_t size)
...
+// Block has format: Size of data area in words | data area in words | Size of data area in words
```
kjbracey-arm: Not all of those "in words" can mean the same thing. How about

  size [1 word] | data area [abs(size) words] | size [1 word]

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -117,146 +114,96 @@ void dev_stat_update(mem_stat_update_t type, int16_t size)
...
-            sector_start -= -(size_end);
-        } else {
-            sector_start -= size_end;
-        }
-        size_start = *sector_start;
+        end -= (1 + abs(size_start));
     }
-    if (size_start == size_end) {
+
+    if (size_start && size_start == *end) {
```
kjbracey-arm: Still don't like this. You've got the nasty boolean + comparison mix, and you're not doing other sanity checks. At a minimum make this "size_start != 0", even if you don't want the other checks.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -117,146 +114,96 @@ void dev_stat_update(mem_stat_update_t type, int16_t size)
...
         ret_val = 0;
     }
     return ret_val;
 }
 #endif
 
-void *ns_dyn_mem_alloc(int16_t alloc_size)
+#ifdef STANDARD_MALLOC
+#ifdef USE_IAR
```
kjbracey-arm: Hmm. Take this out? I don't really believe it can still be required.

Note that it's actually "ifdef STANDARD_MALLOC", so it's not even protecting the real allocator code! When it was originally put in in 2013, it was unconditional or ifndef. Probably working around some error in an earlier version of the real allocator.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -117,146 +114,96 @@ void dev_stat_update(mem_stat_update_t type, int16_t size)
...
-
-    int *ptr = heap_main_end;
     void *retval = 0;
-    int h_size, alloc_sector, s_size;
-    int moved = 0;
-
-    alloc_sector = heap_alloc_internal_check(alloc_size);
-    if (alloc_sector) {
-        h_size = (heap_size / sizeof(int));
+    int alloc_block = heap_alloc_internal_check(alloc_size);
```
kjbracey-arm: data_size rather than alloc_block?

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -117,146 +114,96 @@ void dev_stat_update(mem_stat_update_t type, int16_t size)
...
-                s_size = *ptr;
-                if (s_size == 0) {
-                    heap_failure(NS_DYN_MEM_HEAP_SECTOR_CORRUPTED);
-                    retval = 0;
-                    break;
-                } else if (s_size < 0) {
-                    //Free
-                    if (((-s_size) >= alloc_sector)) {
+            if (ns_block_validate(ptr, direction) == 0) {
+                int block_total_size = *ptr;
```
kjbracey-arm: block_data_size. It's NOT the total size, which would include that two end markers.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -117,146 +114,96 @@ void dev_stat_update(mem_stat_update_t type, int16_t size)
...
-                //Allocated sector
-                moved += (s_size + 2);
-                ptr -= 2; //Skip Both of length fileds
-                ptr -= s_size;
-                if (heap_main == ptr) {
-                    break;
-                }
+
+                moved += (block_total_size + 2);
+                ptr += direction * (block_total_size + 2);
```
kjbracey-arm: Still being inconsistent about whether multiply by direction is good or bad. I don't have a firm view, but stick to one style (compare ns_block_validate).

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -266,14 +213,15 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
 
 
         if (mem_stat_info_ptr) {
-            alloc_size = (alloc_sector * sizeof(int));
+            // If alloc fail needs to collect size of failed alloc,
```
kjbracey-arm: Comment no longer needed.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -290,178 +238,49 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
 #endif
 }
 
+void *ns_dyn_mem_alloc(int16_t alloc_size)
+{
+    return ns_dyn_mem_internal_alloc(alloc_size, -1);
```
kjbracey-arm: Maybe worth a comment, as this is the top-level function people will glance at. "Allocates from end of heap".

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -290,178 +238,49 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
...
-            end_sector += (next_len + 1);
-            size += (next_len + 2);
+    // Theory of operation: Block is always in form | Len | Data | Len |
+    // So we need to check length of previous (if current not heap start)
+    // and next (if current not heap end) blocks. Negative length means
+    // free memory so we can merge freed block with those.
+
+    int *start = cur_block;
+    int *end = cur_block + size + 1;
+    int total_size = size;
```
kjbracey-arm: Same argument against "total" as above. Maybe "data_size" and "merged_data_size"?

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -290,178 +238,49 @@ void *ns_dyn_mem_alloc(int16_t alloc_size)
...
+            end--;
         }
-
-    }
-    //Set new sector start & End length
-    if (prev_sector) {
-        *prev_sector = -(prev_len);
-        *cur_sector = -(prev_len);
     }
+    *cur_block = -total_size;
```
kjbracey-arm: Actually, shouldn't this be -data_size (the passed-in-size), technically? Doesn't really matter, but might be less confusing when dissecting crash dumps.

And would make sense to do this FIRST, before starting the merge. (That would also avoid simplify the compiler's job if making the above change, as it wouldn't need to remember data_size).

Maybe it shouldn't even be part of this call - it should be done outside this call, and this should just be a merge. (Which in turn would not bother with a separate size parameter, it would just load from the pointer).

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -493,25 +312,17 @@ void ns_dyn_mem_free(void *block)
         heap_failure(NS_DYN_MEM_DOUBLE_FREE);
     } else if (ptr < heap_main || ptr >= heap_main_end) {
         heap_failure(NS_DYN_MEM_POINTER_NOT_VALID);
-    } else if ((size < 0) || ((ptr + size) >= heap_main_end)) {
+    } else if ((ptr + size) >= heap_main_end) {
         heap_failure(NS_DYN_MEM_POINTER_NOT_VALID);
     } else {
-        // Validate Sector
-        if (ns_sector_validate(ptr, 1) != 0) {
+        // Validate block
```
kjbracey-arm: Silly comment. Repeats the function name.

source/nsdynmemLIB/nsdynmemLIB.c
AnttiKauppila: Already commented in header

PR #36, opened 2016-05-27
Optimized allocation handling

AnttiKauppila:
Added tracking for memory. 
Now code starts finding in a bit more sophisticated way.
Whenever free block is found during new alloc search it will become smallest/biggest possible start for next alloc. Code could be optimized for 1 step more but wanted to keep code as simple and clear as possible.
@SeppoTakalo @juhhei01 @kjbracey-arm Could you review this change?

### DIFF
--- a/source/nsdynmemLIB/nsdynmemLIB.c
+++ b/source/nsdynmemLIB/nsdynmemLIB.c
@@ -26,6 +26,9 @@ static int *heap_main = 0;
 static int *heap_main_end = 0;
 static uint16_t heap_size = 0;
 
+static int *smallest_possible_start;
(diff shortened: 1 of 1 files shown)


kjbracey-arm:
Don't like it much. It's a kind of cute trick, but is it really simpler than having a more conventional free list, which would work an awful lot better?

I reckon you might be able to achieve that in the same number of lines.

kjbracey-arm:
(And to be honest, I'm not sure I totally understand the logic, whereas the free list concept is very obvious. At least to me.)

AnttiKauppila:
As discussed let's do it with lists

PR #38, opened 2016-06-09
fix yotta builds that don't define FEA_TRACE_SUPPORT

tommikas:
@teetak01 @jupe 

@AnttiKauppila This is an issue in 3.5.2. This and publishing a newer version of mbed-trace should fix it.


teetak01:
+1, should be published ASAP

tommikas:
Turns out this isn't _that_ critical, but should probably still be included.

tommikas:
Jumped the gun a bit. Looks like this isn't necessary.

PR #40, opened 2016-06-22
Add ip6_prefix_tos function.

tommikas:
And ip6tos now returns length of string.

### DIFF
--- a/mbed-client-libservice/ip6string.h
+++ b/mbed-client-libservice/ip6string.h
@@ -18,25 +18,49 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
+
+#include "ns_types.h"
 /**
  * Print binary IPv6 address to a string.
+ *
  * String must contain enough room for full address, 40 bytes exact.
  * IPv4 tunneling addresses are not covered.
+ *
  * \param ip6addr IPv6 address.
- * \p buffer to write string to.
+ * \param p buffer to write string to.
+ * \return length of generated string excluding the terminating null character
  */
-void ip6tos(const void *ip6addr, char *p);
+uint_fast8_t ip6tos(const void *ip6addr, char *p);
+
+/**
+ * Print binary IPv6 prefix to a string.
+ *
+ * String buffer `p` must contain enough room for a full address and prefix length, 44 bytes exact.
+ * Bits in the `prefix` buffer beyond `prefix_len` bits are not shown and only the bytes containing the
+ * prefix bits are read. I.e. for a 20 bit prefix 3 bytes are read, and for a 0 bit prefix 0 bytes are
+ * read (thus if `prefix_len` is zero, `prefix` can be NULL).
+ * `prefix_len` must be 0 to 128.
+ *
+ * \param prefix IPv6 prefix.
+ * \param prefix_len length of `prefix` in bits.
+ * \param p buffer to write string to.
+ * \return length of generated string excluding the terminating null character, or 0 for an error, such as 'prefix_len' > 128
+ */
+uint_fast8_t ip6_prefix_tos(const void *prefix, uint_fast8_t prefix_len, char *p);
 
 /**
  * Convert numeric IPv6 address string to a binary.
+ *
  * IPv4 tunneling addresses are not covered.
+ *
  * \param ip6addr IPv6 address in string format.
  * \param len Lenght of ipv6 string, maximum of 41.
  * \param dest buffer for address. MUST be 16 bytes.
  */
 void stoip6(const char *ip6addr, size_t len, void *dest);
 /**
- * Find out numeric IPv6 address prefix length
+ * Find out numeric IPv6 address prefix length.
+ *
  * \param ip6addr  IPv6 address in string format
  * \return prefix length or 0 if it not given
  */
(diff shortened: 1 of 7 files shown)


tommikas:
@kjbracey-arm @SeppoTakalo @jupe

mbed-client-libservice/ip6string.h
```diff
@@ -18,14 +18,25 @@
...
+#include "ns_types.h"
 /**
  * Print binary IPv6 address to a string.
  * String must contain enough room for full address, 40 bytes exact.
  * IPv4 tunneling addresses are not covered.
  * \param ip6addr IPv6 address.
  * \p buffer to write string to.
  */
-void ip6tos(const void *ip6addr, char *p);
+uint8_t ip6tos(const void *ip6addr, char *p);
```
kjbracey-arm: If you don't want to return size_t (and there are reasons not to on small platforms), use uint_fast8_t instead. Saves bigger platforms messing around trying to makes sure it's narrowed - they can just return 32-bit.

mbed-client-libservice/ip6string.h
```diff
@@ -18,14 +18,25 @@
...
+
+/**
+ * Print binary IPv6 prefix to a string.
+ * String buffer must contain enough room for a full address and prefix length, 44 bytes exact.
+ * \param prefix IPv6 prefix.
+ * \param prefix_len length of prefix in bits.
+ * \p buffer to write string to.
+ */
+uint8_t ip6_prefix_tos(const uint8_t *prefix, uint8_t prefix_len, char *p);
```
kjbracey-arm: While touching this, fix the doxygen "\p" on both functions.

prefix being uint8_t * rather than void * is inconsistent with ip6tos.

Document what the behaviour is with extra bits beyond the prefix length - presumably they're not shown. And presumably extra bytes aren't read? So prefix_len 20 accesses 3 bytes, and prefix_len 0 accesses 0 bytes, making NULL, 0 legal.

Note that prefix length must be 0-128. Again, it count be uint_fast8_t.

source/libip6string/ip6tos.c
```diff
@@ -92,4 +95,27 @@ void ip6tos(const void *ip6addr, char *p)
         }
     }
     *p++ = '\0';
+
+    // Return length of generated string, excluding the terminating null-character
+    return p - p_orig - 1;
```
kjbracey-arm: Could not increment p in the first place. Not that it would affect the compiled code, I guess.

source/libip6string/ip6tos.c
```diff
@@ -92,4 +95,27 @@ void ip6tos(const void *ip6addr, char *p)
...
+}
+
+uint8_t ip6_prefix_tos(const uint8_t *prefix, uint8_t prefix_len, char *p)
+{
+    char *wptr = p;
+    uint8_t addr[16] = {0};
+
+    if (((prefix_len != 0) && (prefix == NULL )) || prefix_len > 128) {
```
kjbracey-arm: Stray space after NULL, and for me this is past the "too much bracketing" threshold.

Isn't this clearer?

    if ((prefix_len != 0 && prefix == NULL) || prefix_len > 128)

Not actually sure if it's worth being defensive here, rather than in the outer layer. ip6tos isn't being defensive, but then I guess there's less to get wrong there.

test/libService/unittest/ip6tos/ip6tos_test.cpp
```diff
@@ -0,0 +1,93 @@
+/*
```
kjbracey-arm: What happened to the original unit tests? There was a quite comprehensive set at one point, covering all the tricky cases. Could you see if you can resurrect them from the history and incorporate?

source/libip6string/ip6tos.c
```diff
@@ -92,4 +95,27 @@ void ip6tos(const void *ip6addr, char *p)
...
+        strcpy(p, "<err>");
+        return 5;
+    }
+
+    // Generate prefix part of the string
+    bitcopy(addr, prefix, prefix_len);
+    wptr += ip6tos(addr, wptr);
+    // Add the prefix length part of the string
+    wptr += sprintf(wptr, "/%u", prefix_len);
```
kjbracey-arm: If you do change to uint_fast8_t, this should be "/%"PRIuFAST8. Technically I guess it should be "/%"PRIu8 now.

tommikas:
Addressed the review comments and added the old tests. Could you review again please.

tommikas:
Fixed up comments a bit and squashed commits a little.

jupe:
Lgtm

mbed-client-libservice/ip6string.h
```diff
@@ -18,25 +18,49 @@
...
 /**
  * Print binary IPv6 address to a string.
+ *
  * String must contain enough room for full address, 40 bytes exact.
  * IPv4 tunneling addresses are not covered.
+ *
  * \param ip6addr IPv6 address.
- * \p buffer to write string to.
+ * \param p buffer to write string to.
+ * \return length of generated string
```
SeppoTakalo: Is the terminating `\0` character included in the length?

SeppoTakalo:
+1

mbed-client-libservice/ip6string.h
tommikas: It isn't. Good point though - I'll modify the comments to document that.

PR #42, opened 2016-07-07
Fix build on GCC older that 4.6.0

SeppoTakalo:
GCC 4.6.0 added support to `#pragma GCC diagnostic push` and others that are used by `ns_types.h`


SeppoTakalo:
@kjbracey-arm Please review.

CC: @tommikas

PR #43, opened 2016-07-15
Remove ns_trace

tommikas:
ns_trace.h simply includes mbed_trace.h while ensuring that FEA_TRACE_SUPPORT is set if HAVE_DEBUG was.

Also removed unit tests for libTrace and replaced ns_trace with mbed_trace for other modules' tests.


tommikas:
@SeppoTakalo @kjbracey-arm Could you please review.

SeppoTakalo:
+1

SeppoTakalo:
Can you restart the tests somehow? I cannot see their failure.

tommikas:
The tests that appear to have failed were yotta builds that Mika removed from the current set of runs (which is why they lead to 404 now).  When I created the PR there were no runs configured for nanostack-libservice because they apparently hadn't been reconfigured after some changes to the repository.

Anyway, the yotta builds would/did fail anyway, because libservice has no direct dependency defined for mbed-trace. This becomes a problem because ns_trace.h includes mbed_trace.h, which of course is nowhere to be found when building the library alone. This hasn't become a problem before because the CI builds weren't there, and because all the applications using libservice also use mbed-trace so it's been available for normal builds including libservice.

This would be a problem however, if an application wants to use libservice without mbed-trace, so perhaps the location (or contents) of ns_trace.h should be reconsidered...

A second problem is that ns_nvm_helper.c has a few trace calls, and since  [this commit](https://github.com/ARMmbed/nanostack-libservice/commit/b728a8ea62b54a16b5daba8b4d781d4dce6f266d) those functions haven't existed within libservice.

tommikas:
@SeppoTakalo Is this ok?

tommikas:
Needed for https://github.com/ARMmbed/mbed-os/pull/4185

mbed-client-libservice/ns_trace.h
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015 ARM Limited. All rights reserved.
+ * Copyright (c) 2015-2016 ARM Limited. All rights reserved.
```
teetak01: Should this be updated to 2017?

tommikas: That would make sense.

Issue #44, opened 2016-08-10
Broken test for definition of INT_FAST24_MAX macro (and isomorphic variants for uint  and other widths).


mjs-arm:
The logic tests the presence of:

INT24_FAST_MAX

and defines the presence of:

INT_FAST24_MAX

They should be the same, the latter is correct.

kjbracey-arm:
Whoops. 

Don't suppose you've actually found a system with real 24-bit types have you? Always seemed like an obvious optimisation thing for at least C99 compilers for 8-bit systems to provide. Guess the implementation cost versus benefit ratio isn't brilliant though.

PR #45, opened 2016-08-10
Correct tests and definitions for 24-bit types

kjbracey-arm:
All tests and some definitions were using backwards forms like
"INT24_FAST_MAX" rather than the correct "INT_FAST24_MAX".

Error has gone unobserved for some time - would only be a real
problem if we encountered a system with these predefined, or
another header attempting to do the same thing correctly.


kjbracey-arm:
@SeppoTakalo, @mikaleppanen, @mjs-arm 

Fixes #44.

mikaleppanen:
+1

mjs-arm:
LGTM

PR #52, opened 2016-12-21
Moved block_data_size to the beginning of function

AnttiKauppila:


kjbracey-arm:
This was a false positive

PR #54, opened 2017-02-09
add cmake file to linux port

netanelgonen:


netanelgonen:
decided ti handle port from higher level

PR #55, opened 2017-02-24
Change all stats to unsigned integer to allow reporting full 64kB of …

SeppoTakalo:
…heap.

### DIFF
--- a/mbed-client-libservice/nsdynmemLIB.h
+++ b/mbed-client-libservice/nsdynmemLIB.h
@@ -47,12 +47,12 @@ typedef enum {
  */
 typedef struct mem_stat_t {
     /*Heap stats*/
-    int16_t heap_sector_size;                   /**< Heap total Sector len. */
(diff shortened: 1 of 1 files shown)


SeppoTakalo:
Actually, If we do anything for this.. should we probably change all types to `size_t` or `ssize_t`?

Feels a bit stupid to limit ourself to 64kB internally on devices with 256kB of RAM.

kjbracey-arm:
This is a legacy "be nice to 8-bit or 16-bit platforms" thing. I think size_t would probably be appropriate.

Conceivably ssize_t might be safer for ones that are already signed - it's possible some sanity-check code checks things going negative on some sort of imbalance.

artokin:
+1, would be nice to get correction merged.
Heap statistics are traced in nanostack_border_router and we get figures like:
```DutThread: [INFO][app ]: Heap size: 50000, Reserved: 49360, Reserved max: 32424, Alloc fail: 254```

artokin:
This change is made in https://github.com/PelionIoT/nanostack-libservice/pull/60, this PR can be closed.

PR #56, opened 2017-03-29
Add more CLZ functions

kjbracey-arm:
Add 16- and 32-bit forms of count leading zeros.


SeppoTakalo:
You need to update Nanostack stubs also. Looks like the compilation now fails beause they are stubbing `common_count_leading_zeros()` with it fails to find the `_8` version.

PR #59, opened 2017-05-23
Add dependency on mbed-trace

bogdanm:
'mbed-client-libservice/ns_trace.h' includes "mbed_trace.h", so this
module needs a dependency on ARMmbed/mbed-trace.

### DIFF
--- a/module.json
+++ b/module.json
@@ -11,6 +11,8 @@
   "extraIncludes": [
     "mbed-client-libservice"
   ],
-  "dependencies": {},
+  "dependencies": {
+      "mbed-trace": "ARMmbed/mbed-trace"
+  },
   "targetDependencies": {}
 }


bogdanm:
@kjbracey-arm

bogdanm:
The CI failure doesn't seem to be related to this PR.

bogdanm:
Ping @kjbracey-arm @AnttiKauppila

kjbracey-arm:
That would be a circular dependency. Is that acceptable?

I note that this library doesn't itself use mbed-trace, except to provide ns_trace.h for its users, so its a slightly weaker dependency than most.

bogdanm:
Those are very good points. Circular dependencies, although undesirable, do work in yotta. In this particular case, I cloned `mbed-trace` and compiled it with a `nanostack-libservice` library that has this PR applied. It compiled without problems.
About this being a weaker dependency, I agree. The problem isn't even with this library itself, because it compiles as it is with yotta. However, when I try to compile `sal-stack-nanostack-eventloop` (which depends on this module), this is what I get:

```
In file included from /home/bogdanm/update/rel1.2/with-yotta/sal-stack-nanostack-eventloop/yotta_modules/nanostack-libservice/source/nvmHelper/ns_nvm_helper.c:9:0:
/home/bogdanm/update/rel1.2/with-yotta/sal-stack-nanostack-eventloop/yotta_modules/nanostack-libservice/mbed-client-libservice/ns_trace.h:56:35: fatal error: mbed-trace/mbed_trace.h: No such file or directory
 #include "mbed-trace/mbed_trace.h"
                                   ^
```

I can fix that by specifying `mbed-trace` as a dependency of `sal-stack-nanostack-eventloop` itself, but I don't think that's the right thing to do for two reasons:

1, `nanostack-libservice` does include `mbed-trace.h`. So, if we were to play by the rules, this means that it needs to depend on the `mbed-trace` library as far as yotta is concerned.
2. I'd rather add a single dependency on `mbed-trace` for `nanostack-libservice` than add dependencies on `mbed-trace` for everything that depends on `nanostack-libservice` and can't compile because it doesn't depend on `mbed-trace` itself.

bogdanm:
So, are you convinced yet ? :smile:

PR #60, opened 2017-06-21
Dynamic memory options

tpihlaja:
Reviewing object oriented version dynmemLIB

### DIFF
--- a/mbed-client-libservice/nsdynmemLIB.h
+++ b/mbed-client-libservice/nsdynmemLIB.h
@@ -18,8 +18,11 @@
 /**
  * \file nsdynmemLIB.h
  * \brief Dynamical Memory API for library model
- *
+ * nsdynmemlib provides access to one default heap, along with the ability to use extra user heaps.
+ * ns_dyn_mem_alloc/free always access the default heap initialised by ns_dyn_mem_init.
+ * ns_mem_alloc/free access a user heap initialised by ns_mem_init. User heaps are identified by a book-keeping pointer.
  */
+
 #ifndef NSDYNMEMLIB_H_
 #define NSDYNMEMLIB_H_
 #ifdef __cplusplus
@@ -27,6 +30,12 @@ extern "C" {
 #endif
 
 #include "ns_types.h"
+
+// Added to maintain backward compatibility with older implementation of ns_dyn_mem APIs
+#define NSDYNMEMLIB_API_VERSION 2
+
+typedef uint16_t ns_mem_block_size_t; //external interface unsigned heap block size type
+typedef uint16_t ns_mem_heap_size_t; //total heap size type.
 
 /*!
  * \enum heap_fail_t
@@ -47,13 +56,16 @@ typedef enum {
  */
 typedef struct mem_stat_t {
     /*Heap stats*/
-    int16_t heap_sector_size;                   /**< Heap total Sector len. */
-    int16_t heap_sector_alloc_cnt;              /**< Reserved Heap sector cnt. */
-    int16_t heap_sector_allocated_bytes;        /**< Reserved Heap data in bytes. */
-    int16_t heap_sector_allocated_bytes_max;    /**< Reserved Heap data in bytes max value. */
+    ns_mem_heap_size_t heap_sector_size;                   /**< Heap total Sector len. */
+    ns_mem_heap_size_t heap_sector_alloc_cnt;              /**< Reserved Heap sector cnt. */
+    ns_mem_heap_size_t heap_sector_allocated_bytes;        /**< Reserved Heap data in bytes. */
+    ns_mem_heap_size_t heap_sector_allocated_bytes_max;    /**< Reserved Heap data in bytes max value. */
     uint32_t heap_alloc_total_bytes;            /**< Total Heap allocated bytes. */
     uint32_t heap_alloc_fail_cnt;               /**< Counter for Heap allocation fail. */
 } mem_stat_t;
+
+
+typedef struct ns_mem_book ns_mem_book_t;
 
 /**
   * \brief Init and set Dynamical heap pointer and length.
@@ -62,7 +74,7 @@ typedef struct mem_stat_t {
   * \param heap_size size of the heap buffer
   * \return None
   */
-extern void ns_dyn_mem_init(uint8_t *heap, uint16_t h_size, void (*passed_fptr)(heap_fail_t), mem_stat_t *info_ptr);
+extern void ns_dyn_mem_init(void *heap, ns_mem_heap_size_t h_size, void (*passed_fptr)(heap_fail_t), mem_stat_t *info_ptr);
 
 
 /**
@@ -84,7 +96,7 @@ extern void ns_dyn_mem_free(void *heap_ptr);
   * \return 0, Allocate Fail
   * \return >0, Pointer to allocated data sector.
   */
-extern void *ns_dyn_mem_temporary_alloc(int16_t alloc_size);
+extern void *ns_dyn_mem_temporary_alloc(ns_mem_block_size_t alloc_size);
 /**
   * \brief Allocate long period data.
   *
@@ -95,7 +107,7 @@ extern void *ns_dyn_mem_temporary_alloc(int16_t alloc_size);
   * \return 0, Allocate Fail
   * \return >0, Pointer to allocated data sector.
   */
-extern void *ns_dyn_mem_alloc(int16_t alloc_size);
+extern void *ns_dyn_mem_alloc(ns_mem_block_size_t alloc_size);
 
 /**
   * \brief Get pointer to the current mem_stat_t set via ns_dyn_mem_init.
@@ -109,6 +121,65 @@ extern void *ns_dyn_mem_alloc(int16_t alloc_size);
   * \return !=0, Pointer to mem_stat_t.
   */
 extern const mem_stat_t *ns_dyn_mem_get_mem_stat(void);
+
+/**
+  * \brief Init and set Dynamical heap pointer and length.
+  *
+  * \param heap_ptr Pointer to dynamically heap buffer
+  * \param heap_size size of the heap buffer
+  * \return !=0, Pointer to ns_mem_book_t.
+  */
+extern ns_mem_book_t *ns_mem_init(void *heap, ns_mem_heap_size_t h_size, void (*passed_fptr)(heap_fail_t), mem_stat_t *info_ptr);
+
+/**
+  * \brief Free allocated memory.
+  *
+  * \param book Address of book keeping structure
+  * \param heap_ptr Pointer to allocated memory
+  *
+  * \return 0, Free OK
+  * \return <0, Free Fail
+  */
+extern void ns_mem_free(ns_mem_book_t *book, void *heap_ptr);
+/**
+  * \brief Allocate temporary data.
+  *
+  * Space allocate started from beginning of the heap sector
+  *
+  * \param book Address of book keeping structure
+  * \param alloc_size Allocated data size
+  *
+  * \return 0, Allocate Fail
+  * \return >0, Pointer to allocated data sector.
+  */
+extern void *ns_mem_temporary_alloc(ns_mem_book_t *book, ns_mem_block_size_t alloc_size);
+/**
+  * \brief Allocate long period data.
+  *
+  * Space allocate started from end of the heap sector
+  *
+  * \param book Address of book keeping structure
+  * \param alloc_size Allocated data size
+  *
+  * \return 0, Allocate Fail
+  * \return >0, Pointer to allocated data sector.
+  */
+extern void *ns_mem_alloc(ns_mem_book_t *book, ns_mem_block_size_t alloc_size);
+
+/**
+  * \brief Get pointer to the current mem_stat_t set via ns_mem_init.
+  *
+  * Get pointer to the statistics information, if one is set during the
+  * initialization. This may be useful for statistics collection purposes.
+  *
+  * Note: the caller may not modify the returned structure.
+  *
+  * \param book Address of book keeping structure
+  *
+  * \return NULL, no mem_stat_t was given on initialization
+  * \return !=0, Pointer to mem_stat_t.
+  */
+extern const mem_stat_t *ns_mem_get_mem_stat(ns_mem_book_t *book);
 
 #ifdef __cplusplus
 }
--- a/source/nsdynmemLIB/nsdynmemLIB.c
+++ b/source/nsdynmemLIB/nsdynmemLIB.c
@@ -20,102 +20,128 @@
 #include <stdlib.h>
 #include "ns_list.h"
 
-void (*heap_failure_callback)(heap_fail_t);
-
 #ifndef STANDARD_MALLOC
-static int *heap_main = 0;
-static int *heap_main_end = 0;
-static uint16_t heap_size = 0;
-
 typedef enum mem_stat_update_t {
     DEV_HEAP_ALLOC_OK,
     DEV_HEAP_ALLOC_FAIL,
     DEV_HEAP_FREE,
 } mem_stat_update_t;
 
-
-static mem_stat_t *mem_stat_info_ptr = 0;
-
 typedef struct {
     ns_list_link_t link;
 } hole_t;
 
-static NS_LIST_DEFINE(holes_list, hole_t, link);
+typedef int ns_mem_word_size_t; // internal signed heap block size type
+
+/* struct for book keeping variables */
+struct ns_mem_book {
+    ns_mem_word_size_t     *heap_main;
+    ns_mem_word_size_t     *heap_main_end;
+    mem_stat_t *mem_stat_info_ptr;
+    void (*heap_failure_callback)(heap_fail_t);
+    NS_LIST_HEAD(hole_t, link) holes_list;
+    ns_mem_heap_size_t heap_size;
+};
+
+static ns_mem_book_t *default_book; // heap pointer for original "ns_" API use
 
 // size of a hole_t in our word units
-#define HOLE_T_SIZE ((sizeof(hole_t) + sizeof(int) - 1) / sizeof(int))
+#define HOLE_T_SIZE ((sizeof(hole_t) + sizeof(ns_mem_word_size_t) - 1) / sizeof(ns_mem_word_size_t))
 
-static NS_INLINE hole_t *hole_from_block_start(int *start)
+static NS_INLINE hole_t *hole_from_block_start(ns_mem_word_size_t *start)
 {
     return (hole_t *)(start + 1);
 }
 
-static NS_INLINE int *block_start_from_hole(hole_t *start)
+static NS_INLINE ns_mem_word_size_t *block_start_from_hole(hole_t *start)
 {
-    return ((int *)start) - 1;
+    return ((ns_mem_word_size_t *)start) - 1;
 }
 
-
-static void heap_failure(heap_fail_t reason)
+static void heap_failure(ns_mem_book_t *book, heap_fail_t reason)
 {
-    if (heap_failure_callback) {
-        heap_failure_callback(reason);
+    if (book->heap_failure_callback) {
+        book->heap_failure_callback(reason);
     }
 }
 
 #endif
 
-void ns_dyn_mem_init(uint8_t *heap, uint16_t h_size, void (*passed_fptr)(heap_fail_t), mem_stat_t *info_ptr)
+void ns_dyn_mem_init(void *heap, ns_mem_heap_size_t h_size,
+                     void (*passed_fptr)(heap_fail_t), mem_stat_t *info_ptr)
 {
-#ifndef STANDARD_MALLOC
-    int *ptr;
-    int temp_int;
-    /* Do memory alignment */
-    temp_int = ((uintptr_t)heap % sizeof(int));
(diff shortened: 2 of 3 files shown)


kjbracey-arm:
Looks good.

The main thing I want to see here is one more commit that removes the implementation from nsdynmemlib.c, and makes it just call this new code with a static context.

Partly because it's important for the final result to keep the code size down, but mainly because that will be the best way to the new implementation now - we've got a large body of code using nsdynmemlib.h, and if that gets routed to this, this gets hella testing.

Are there any significant blockers to that?

I'll add a couple of minor notes to the code body, and at some point I'll need to run a diff between the current nsdynmemlib.c and this C file to check no changes have got lost.

tpihlaja:
I imagined implementations would be merged so that both interfaces would be available. I made some experiment already. There I combined the modules and added "_oo_" to new function names to make them different symbols. Afterwards code could be internally merged and optimized for code size. But maybe it would be possible with some macro to just route the existing calls to new implementation (with static context)).

mbed-client-libservice/m2mdynmemLIB.h
```diff
@@ -0,0 +1,124 @@
...
+ * \enum heap_fail_t
+ * \brief Dynamically heap system failure call back event types.
+ */
+typedef enum {
+    M2M_DYN_MEM_NULL_FREE,               /**< ns_dyn_mem_free(), NULL pointer free [obsolete - no longer faulted]  */
+    M2M_DYN_MEM_DOUBLE_FREE,                     /**< ns_dyn_mem_free(), Possible double pointer free */
+    M2M_DYN_MEM_ALLOCATE_SIZE_NOT_VALID, /**< Allocate size is 0 or smaller or size is bigger than max heap size  */
+    M2M_DYN_MEM_POINTER_NOT_VALID,       /**< ns_dyn_mem_free(), try to free pointer which not at heap sector */
+    M2M_DYN_MEM_HEAP_SECTOR_CORRUPTED,   /**< Heap system detect sector corruption */
+    M2M_DYN_MEM_HEAP_SECTOR_UNITIALIZED /**< ns_dyn_mem_free(), ns_dyn_mem_temporary_alloc() or ns_dyn_mem_alloc() called before ns_dyn_mem_init() */
```
kjbracey-arm: Lingering ns_dyn references in the comments. Maybe take the opportunity to line up the comments (we can ignore spaces when diffing)

mbed-client-libservice/m2mdynmemLIB.h
```diff
@@ -0,0 +1,124 @@
...
+} m2m_mem_stat_t;
+
+/**
+  * \brief Init and set Dynamical heap pointer and length.
+  *
+  * \param heap_ptr Pointer to dynamically heap buffer
+  * \param heap_size size of the heap buffer
+  * \return None
+  */
+extern void m2m_dyn_mem_init(uint8_t *heap, m2m_heap_size_t h_size, void (*passed_fptr)(m2m_heap_fail_t), m2m_mem_stat_t *info_ptr);
```
kjbracey-arm: Can you make the heap pointers be void *? Just to be a bit nicer, and would eliminate some casts in the C file, I think.

kjbracey-arm:
I would just suggest making it like:

    void *nsdynmemlib_temporary_alloc(uint16_t size)
    {
        return m2mdynmemlib_temporary_alloc(&ns_heap, size);
    }

That's going to be a bit better from a code size point of view than using a macro so everyone calls with 2 parameters.

mbed-client-libservice/m2mdynmemLIB.h
```diff
@@ -0,0 +1,124 @@
...
+ * \brief Dynamical Memory API for library model
+ *
+ */
+#ifndef M2MDYNMEMLIB_H_
+#define M2MDYNMEMLIB_H_
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "mbed-client-libservice/ns_types.h"
```
tpihlaja: Yes. I had done this change in some test environment locally.

tpihlaja:
Looks good and simpler than a macro.

tpihlaja:
One worrying thing with the new implementation is the alignment. I thought about returning possibly aligned pointer from init function, but maybe, if alignment is rarely needed, alignment check could be made an assertion instead of changing pointer? Possibly alignment code could put behind compiler flags to be removed completely after initialisation code is tested in an environment. In effect we would trust the developer to always give properly aligned address to init function.

kjbracey-arm:
Good point. I would suggest revising the API so that you pass void * to the init, then it returns an (opaque) aligned book_t *. Then users pass that to the other functions, rather than the raw heap pointer.

kjbracey-arm:
I guess that is fractionally less efficient, as it's one more load per heap operation to load from the pointer. But it's just the same as the original ns_dyn_mem_init.

The alternative would be a check in init that the pointer is properly aligned, but I think I like passing a proper type to the non-init functions for type-checking.

mbed-client-libservice/m2mdynmemLIB.h
tpihlaja: Fixed.

mbed-client-libservice/m2mdynmemLIB.h
tpihlaja: Fixed.

mbed-client-libservice/m2mdynmemLIB.h
tpihlaja: Should it be changed for ns_dyn_mem_init also?

tpihlaja: nsdynmemLIB.h is kept untouched, but otherwise changes are done.

mbed-client-libservice/m2mdynmemLIB.h
```diff
@@ -0,0 +1,128 @@
...
+    /*Heap stats*/
+    int16_t heap_sector_size;                /**< Heap total Sector len. */
+    int16_t heap_sector_alloc_cnt;                  /**< Reserved Heap sector cnt. */
+    int16_t heap_sector_allocated_bytes;     /**< Reserved Heap data in bytes. */
+    int16_t heap_sector_allocated_bytes_max; /**< Reserved Heap data in bytes max value. */
+    uint32_t heap_alloc_total_bytes;                 /**< Total Heap allocated bytes. */
+    uint32_t heap_alloc_fail_cnt;                    /**< Counter for Heap allocation fail. */
+} m2m_mem_stat_t;
+
+struct book;
```
kjbracey-arm: Now this is public API it needs a prefix matching the rest. And it's better to have it as a public typedef (while keeping the actual structure hidden).

On that note not massively fond of the m2m naming in libservice. We do need to name it differently somehow for compatibility. Only other thing I can think of is just a subtle change like dropping the pointless "dyn" - "nsmemlib.h". Not that the "ns" prefix is brilliant either, but matches the rest of the surrounding code.

But I could live with m2mdynmemlib.

tpihlaja: Added a prefix and typedef.

yogpan01:
@kjbracey-arm Does this look good now ? Can it be merged ?

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -20,102 +20,132 @@
...
+{
+    static_context = ns_mem_init((void *)heap, h_size,
+                                      (void (*)(heap_fail_t))passed_fptr,
+                                                   (mem_stat_t *)info_ptr);
+}
+
+const mem_stat_t *ns_dyn_mem_get_mem_stat(void)
+{
+#ifndef STANDARD_MALLOC
+    return (mem_stat_t *)ns_mem_get_mem_stat(static_context);
```
kjbracey-arm: Cast no longer needed

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -20,102 +20,132 @@
...
-static NS_INLINE int *block_start_from_hole(hole_t *start)
+static NS_INLINE ns_mem_word_size_t *block_start_from_hole(hole_t *start)
 {
-    return ((int *)start) - 1;
+    return ((ns_mem_word_size_t *)start) - 1;
 }
 
-
-static void heap_failure(heap_fail_t reason)
+static void heap_failure(void (*callback)(heap_fail_t), heap_fail_t reason)
```
kjbracey-arm: Neater and more efficient to pass the book pointer here, rather than the function pointer.

mbed-client-libservice/nsdynmemLIB.h
```diff
@@ -55,6 +55,12 @@ typedef struct mem_stat_t {
     uint32_t heap_alloc_fail_cnt;               /**< Counter for Heap allocation fail. */
 } mem_stat_t;
 
+
+typedef uint16_t ns_mem_block_size_t; //external interface unsigned heap block size type
```
kjbracey-arm: You can make the ` int16_t` s in the stats be ` ns_mem_heap_size_t`s - picks up that other pull request, keeps binary compatibility for current builds, and gives room for future expansion.

mbed-client-libservice/nsdynmemLIB.h
```diff
@@ -110,6 +116,65 @@ extern void *ns_dyn_mem_alloc(int16_t alloc_size);
   */
 extern const mem_stat_t *ns_dyn_mem_get_mem_stat(void);
 
+/*
+  * \brief Init and set Dynamical heap pointer and length.
+  *
+  * \param heap_ptr Pointer to dynamically heap buffer
+  * \param heap_size size of the heap buffer
+  * \return None
```
kjbracey-arm: There is a return now.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -20,102 +20,132 @@
...
+typedef struct ns_mem_book {
+    ns_mem_word_size_t     *heap_main;
+    ns_mem_word_size_t     *heap_main_end;
+    mem_stat_t *mem_stat_info_ptr;
+    void (*heap_failure_callback)(heap_fail_t);
+    NS_LIST_HEAD(hole_t, link) holes_list;
+    ns_mem_heap_size_t heap_size;
+} ns_mem_book_t;
+
+static ns_mem_book_t *static_context; // heap pointer for original "ns_" API use
```
kjbracey-arm: Would `default_context` be better? Or `default_heap` or `default_book`? Don't think "static" is meaningful here.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -20,102 +20,132 @@
...
+    temp_int = (h_size % sizeof(ns_mem_word_size_t));
     if (temp_int) {
-        h_size -= (sizeof(int) - temp_int);
+        h_size -= (sizeof(ns_mem_word_size_t) - temp_int);
     }
-    heap_main = (int *)heap; // SET Heap Pointer
-    heap_size = h_size; //Set Heap Size
-    temp_int = (h_size / sizeof(int));
+    book = (ns_mem_book_t *)heap;
+    book->holes_list.slist.first_entry = NULL;
```
kjbracey-arm: Um, no. This is ns_list_init hard-coded, which you've already got below. I guess the original code both statically initialised it with NS_LIST_DEFINE, then dynamically with the ns_list_init, which is how you've ended up doing it twice.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -20,102 +20,132 @@
...
+
+    ns_mem_word_size_t *ptr;
+    ns_mem_word_size_t temp_int;
     /* Do memory alignment */
-    temp_int = ((uintptr_t)heap % sizeof(int));
+    temp_int = ((uintptr_t)heap % sizeof(ns_mem_word_size_t));
     if (temp_int) {
-        heap += (sizeof(int) - temp_int);
-        h_size -= (sizeof(int) - temp_int);
+        heap += (sizeof(ns_mem_word_size_t) - temp_int);
```
kjbracey-arm: This addition (to void *) isn't legal in standard C. It's a GCC extension.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -262,18 +299,28 @@ static void *ns_dyn_mem_internal_alloc(const int16_t alloc_size, int direction)
...
+
+void *ns_mem_temporary_alloc(ns_mem_book_t *heap, ns_mem_block_size_t alloc_size)
+{
+    return ns_mem_internal_alloc(heap, alloc_size, 1);
+}
+
 void *ns_dyn_mem_alloc(int16_t alloc_size)
 {
-    return ns_dyn_mem_internal_alloc(alloc_size, -1);
+    return ns_mem_alloc(static_context, (ns_mem_block_size_t)alloc_size);
```
kjbracey-arm: It would be better to change ns_dyn_mem functions to take the same ns_mem_block_size_t types, and drop the casts. (Not a functional change, as they're currently the same types).

source/nsdynmemLIB/nsdynmemLIB.c
yogpan01: ok, done.

source/nsdynmemLIB/nsdynmemLIB.c
yogpan01: ok, done.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -20,102 +20,132 @@
...
-    temp_int = (h_size % sizeof(int));
+    temp_int = (h_size % sizeof(ns_mem_word_size_t));
     if (temp_int) {
-        h_size -= (sizeof(int) - temp_int);
+        h_size -= (sizeof(ns_mem_word_size_t) - temp_int);
     }
-    heap_main = (int *)heap; // SET Heap Pointer
-    heap_size = h_size; //Set Heap Size
-    temp_int = (h_size / sizeof(int));
+    book = (ns_mem_book_t *)heap;
```
kjbracey-arm: Cast no longer required, because it's `void *`.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -20,102 +20,132 @@
...
     }
 }
 
 #endif
 
-void ns_dyn_mem_init(uint8_t *heap, uint16_t h_size, void (*passed_fptr)(heap_fail_t), mem_stat_t *info_ptr)
+void ns_dyn_mem_init(uint8_t *heap, uint16_t h_size,
+                     void (*passed_fptr)(heap_fail_t), mem_stat_t *info_ptr)
+{
+    static_context = ns_mem_init((void *)heap, h_size,
```
kjbracey-arm: Cast not required.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -20,102 +20,132 @@
...
+                                      (void (*)(heap_fail_t))passed_fptr,
+                                                   (mem_stat_t *)info_ptr);
```
kjbracey-arm: Cast not required - it's now the same thing.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -20,102 +20,132 @@
...
 }
 
 #endif
 
-void ns_dyn_mem_init(uint8_t *heap, uint16_t h_size, void (*passed_fptr)(heap_fail_t), mem_stat_t *info_ptr)
+void ns_dyn_mem_init(uint8_t *heap, uint16_t h_size,
+                     void (*passed_fptr)(heap_fail_t), mem_stat_t *info_ptr)
+{
+    static_context = ns_mem_init((void *)heap, h_size,
+                                      (void (*)(heap_fail_t))passed_fptr,
```
kjbracey-arm: Cast not required - it's now the same thing.

source/nsdynmemLIB/nsdynmemLIB.c
kjbracey-arm: Suggest `heap = (uint8_t *) heap + (sizeof(ns_mem_word_size_t) - temp_int)`.

mbed-client-libservice/nsdynmemLIB.h
kjbracey-arm: Also update the function prototypes to be heap/block sizes.

yogpan01:
@kjbracey-arm All comments are now fixed, hopefully I have not missed anything :)

mbed-client-libservice/nsdynmemLIB.h
```diff
@@ -47,22 +53,25 @@ typedef enum {
  */
 typedef struct mem_stat_t {
     /*Heap stats*/
-    int16_t heap_sector_size;                   /**< Heap total Sector len. */
-    int16_t heap_sector_alloc_cnt;              /**< Reserved Heap sector cnt. */
-    int16_t heap_sector_allocated_bytes;        /**< Reserved Heap data in bytes. */
-    int16_t heap_sector_allocated_bytes_max;    /**< Reserved Heap data in bytes max value. */
+    ns_mem_block_size_t heap_sector_size;                   /**< Heap total Sector len. */
```
kjbracey-arm: These should be heap sizes, rather than block sizes.

mbed-client-libservice/nsdynmemLIB.h
```diff
@@ -47,22 +53,25 @@ typedef enum {
...
+
 /**
   * \brief Init and set Dynamical heap pointer and length.
   *
   * \param heap_ptr Pointer to dynamically heap buffer
   * \param heap_size size of the heap buffer
   * \return None
   */
-extern void ns_dyn_mem_init(uint8_t *heap, uint16_t h_size, void (*passed_fptr)(heap_fail_t), mem_stat_t *info_ptr);
+extern void ns_dyn_mem_init(uint8_t *heap, ns_mem_heap_size_t h_size, void (*passed_fptr)(heap_fail_t), mem_stat_t *info_ptr);
```
kjbracey-arm: heap to void * here.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -67,11 +67,11 @@ static void heap_failure(ns_mem_book_t *book, heap_fail_t reason)
...
 #endif
 
-void ns_dyn_mem_init(uint8_t *heap, uint16_t h_size,
+void ns_dyn_mem_init(uint8_t *heap, ns_mem_heap_size_t h_size,
                      void (*passed_fptr)(heap_fail_t), mem_stat_t *info_ptr)
 {
-    default_book = ns_mem_init((void *)heap, h_size,
-                                      (void (*)(heap_fail_t))passed_fptr,
+    default_book = ns_mem_init(heap, h_size,
+                               (void (*)(heap_fail_t))passed_fptr,
```
kjbracey-arm: Two unneeded casts still here and next line

kjbracey-arm:
(And the test is pointing out you haven't updated the C file ns_dyn_mem prototypes to match the header file changes)

mbed-client-libservice/nsdynmemLIB.h
```diff
@@ -28,6 +31,9 @@ extern "C" {
 
 #include "ns_types.h"
 
+typedef uint16_t ns_mem_block_size_t; //external interface unsigned heap block size type
```
kjbracey-arm: Please define NSDYNMEMLIB_API_VERSION to 2 so API changes can be detected.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -20,102 +20,130 @@
...
 }
 
 #endif
 
-void ns_dyn_mem_init(uint8_t *heap, uint16_t h_size, void (*passed_fptr)(heap_fail_t), mem_stat_t *info_ptr)
+void ns_dyn_mem_init(void *heap, ns_mem_heap_size_t h_size,
+                     void (*passed_fptr)(heap_fail_t), mem_stat_t *info_ptr)
+{
+    default_book = ns_mem_init(heap, h_size,
+                               (void (*)(heap_fail_t))passed_fptr,
```
kjbracey-arm: Still have the two unnecessary casts. This should just read

    default_book = ns_mem_init(heap, h_size, passed_fptr, info_ptr);

kjbracey-arm:
PRs merged into Nanostack and nanomesh-applications.

kjbracey-arm:
Note that the PR updates for the other repos only cover the old functions.

If you're going to start testing mbed-client code using the new functions in the simulator, we'll need a Linux implementation (the simulator doesn't (usually?) use the nsdynmemLIB.c file for the implementation, but instead has a tracing veneer onto the normal Linux application heap).

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -360,39 +403,45 @@ static void ns_free_and_merge_with_adjacent_blocks(int * const cur_block, int da
 }
 #endif
 
-void ns_dyn_mem_free(void *block)
+void ns_mem_free(ns_mem_book_t *book, void *block)
 {
 #ifndef STANDARD_MALLOC
-    int *ptr = block;
-    int size;
+    if(!book) {
```
kjbracey-arm: Oh, one little request - take this null test out. Let it crash if they pass null book.

(null block is legal, and supposed to be a no-op, but book must be valid).

Issue #61, opened 2017-09-20
Comparison between signed and unsigned integer expressions


korjaa:
https://github.com/ARMmbed/nanostack-libservice/blob/854505d0191e428c07eb25fc9630dc05b21d5a47/source/nsdynmemLIB/nsdynmemLIB.c#L240

https://github.com/ARMmbed/nanostack-libservice/blob/854505d0191e428c07eb25fc9630dc05b21d5a47/source/nsdynmemLIB/nsdynmemLIB.c#L382

korjaa:
Getting this with
`mbed compile -m K64F -t GCC_ARM --profile=mbed-os/tools/profiles/debug.json`

ciarmcom:
ARM Internal Ref: ONME-3201

kjbracey-arm:
Fixed.

PR #62, opened 2017-10-04
Correct some types, solving warning

kjbracey-arm:
GCC is generating warnings about comparison between signed and unsigned
values - https://github.com/ARMmbed/nanostack-libservice/issues/61

This was due to HOLE_T_SIZE being inappropriately of type size_t.
Correct that, and a couple of other stray uses of int and size_t instead
of ns_mem_word_size_t, which would cause problems if the typedef was
changed from int.


kjbracey-arm:
Rebased, Seppo already pushed part of the change. Now just other minor fixes.

PR #64, opened 2017-10-25
Add ARMC6 flags for alignas

artokin:
Add ARMC6 version check to select proper alignas.

### DIFF
--- a/mbed-client-libservice/ns_types.h
+++ b/mbed-client-libservice/ns_types.h
@@ -117,7 +117,7 @@ typedef int_fast32_t int_fast24_t;
 
 /* C11's "alignas" macro, emulated for integer expressions if necessary */
 #ifndef __alignas_is_defined
-#if defined __CC_ARM || defined __TASKING__
+#if defined __CC_ARM || defined __TASKING__ || ( defined(__ARMCC_VERSION) && __ARMCC_VERSION >= 6000000 )
 #define alignas(n) __align(n)
 #define __alignas_is_defined 1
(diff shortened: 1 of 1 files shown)


artokin:
@kjbracey-arm , @SeppoTakalo , @mikter would you please review?

mbed-client-libservice/ns_types.h
```diff
@@ -117,7 +117,7 @@ typedef int_fast32_t int_fast24_t;
 
 /* C11's "alignas" macro, emulated for integer expressions if necessary */
 #ifndef __alignas_is_defined
-#if defined __CC_ARM || defined __TASKING__
+#if defined __CC_ARM || defined __TASKING__ || ( defined(__ARMCC_VERSION) && __ARMCC_VERSION >= 6000000 )
```
kjbracey-arm: So ARM C 6 doesn't define __CC_ARM? But ARM C 5 does define __ARMCC_VERSION, doesn't it, so you could just check for that?

kjbracey-arm: Or check for `__clang__` - would make more sense as you're looking for a language front-end feature.
Although does it actually support __align? Can't see it in documentation.

kjbracey-arm: Whereas it does support the GNU C attribute, and it would pick it up without this change, as `__GNUC__` is defined.

artokin:
Closing this PR as this change is not anymore needed when toolchain is configured properly (in https://github.com/ARMmbed/nanomesh-libservice-root-private/pull/60).

PR #68, opened 2018-06-18
Update max heap size to use size_t

artokin:
-allow >64kB heap allocation by adjusting typedef
-Update memory stats structure to use size_t instead of uint16_t
-Remove excess parameter from function


artokin:
@kjbracey-arm , @SeppoTakalo , @terhei

Issue #71, opened 2018-06-29
New "is this an IPv4/IPv6 literal?" helper


Taiki-San:
In a different PR (https://github.com/ARMmbed/mbed-os/pull/6293#issuecomment-396533529), it was mentioned it'd be helpful to add a new helper function, to go with the new IPv4 functions introduced in https://github.com/ARMmbed/nanostack-libservice/pull/69. I was wondering what shape would you see it take.
I was thinking about adding an `enum` with `IPv4`, `IPv6`, `unknown`, `error` options and a small helper taking in a string and returning either of them. Not sure about the calling convention for the enums, specifically if the project had guidelines.
Was that what @kjbracey-arm had in mind?

ciarmcom:
ARM Internal Ref: ONME-3740

kjbracey-arm:
I think having conversion functions that return a bool like `stoip4` does is sufficient. If `stoip6` did the same, then we can just try each conversion function in turn.

Taiki-San:
Alright, this will require some changes to `stoip6` but the minimal signature change will hopefully not break any third party code.  
Should we still add an helper that calls both functions back to back, or is it too risky, considering the caller shall provide an output buffer large enough for an IPv6 even if it only expects an IPv4?

kjbracey-arm:
Don't see any need for a helper at this layer. You'd then need to invent a "type" enum at this layer, which probably would then need to be converted to the next layer up.

In mbed OS, effectively SocketAddress acts as the helper.

Taiki-San:
Alright, I'm having a look at how to properly return an error in `stoip6`.

PR #72, opened 2018-06-29
Make stoip6 return whether the conversion succeed

Taiki-San:
Implement the change suggested by @kjbracey-arm in https://github.com/ARMmbed/nanostack-libservice/issues/71.
`stoip6` now returns false if the string is too long, if characters besides hexadecimal characters or ':' are detected, or if it's missing fields.

### DIFF
--- a/source/libip6string/stoip6.c
+++ b/source/libip6string/stoip6.c
@@ -20,6 +20,7 @@
 #include "ip6string.h"
 
 static uint16_t hex(const char *p);
+static bool is_hex(char c);
 
 /**
  * Convert numeric IPv6 address string to a binary.
@@ -27,8 +28,9 @@ static uint16_t hex(const char *p);
  * \param ip6addr IPv6 address in string format.
  * \param len Length of ipv6 string.
  * \param dest buffer for address. MUST be 16 bytes.
+ * \return boolean set to true if conversion succeed, false if it didn't
  */
-void stoip6(const char *ip6addr, size_t len, void *dest)
+bool stoip6(const char *ip6addr, size_t len, void *dest)
 {
     uint8_t *addr;
     const char *p, *q;
@@ -37,23 +39,39 @@ void stoip6(const char *ip6addr, size_t len, void *dest)
     addr = dest;
 
     if (len > 39) { // Too long, not possible. We do not support IPv4-mapped IPv6 addresses
-        return;
+        goto error;
     }
 
     // First go forward the string, until end, noting :: position if any
-    for (field_no = 0, p = ip6addr; (len > (size_t)(p - ip6addr)) && *p && field_no < 8; p = q + 1) {
-        q = p;
-        // Seek for ':' or end
-        while (*q && (*q != ':')) {
-            q++;
+    // We're decrementing `len` as we go forward, and stop when it reaches 0
+    for (field_no = 0, p = ip6addr; len && *p; p = q + 1, len -= 1) {
+
+        for (q = p; len && *q && (*q != ':'); len -= 1) { // Seek for ':' or end
+            if (!is_hex(*q++)) { // There must only be hex characters besides ':'
+                goto error;
+            }
         }
-        //Convert and write this part, (high-endian AKA network byte order)
+
+        if ((q - p) > 4) { // We can't have more than 4 hex digits per segment
+            goto error;
+        }
+
+        if (field_no == 8) { // If the address goes farther than 8 segments
+            goto error;
+        }
+
+        // Convert and write this part, (high-endian AKA network byte order)
         addr = common_write_16_bit(hex(p), addr);
         field_no++;
-        //Check if we reached "::"
-        if ((len > (size_t)(q - ip6addr)) && *q && (q[0] == ':') && (q[1] == ':')) {
+
+        // Check if we reached "::"
+        if (len && (q[0] == ':') && (q[1] == ':')) {
+            if (coloncolon != -1) { // We are not supposed to see "::" more than once per address
+                goto error;
+            }
             coloncolon = field_no;
             q++;
+            len--;
         }
     }
 
@@ -65,11 +83,18 @@ void stoip6(const char *ip6addr, size_t len, void *dest)
         addr = dest;
         memmove(addr + head_size + inserted_size, addr + head_size, tail_size);
         memset(addr + head_size, 0, inserted_size);
-    } else if (field_no != 8) {
-        /* Should really report an error if we didn't get 8 fields */
-        memset(addr, 0, 16 - field_no * 2);
+    } else if (field_no != 8) { // Report an error if we didn't get 8 fields
+        goto error;
     }
+    return true;
+
+error:
+    // Fill the output buffer with 0 so we stick to the old failure behavior.
+    // We are however more agressive and wipe the entire address, and do so more often.
+    memset(dest, 0, 16);
+    return false;
 }
+
 unsigned char  sipv6_prefixlength(const char *ip6addr)
 {
     char *ptr = strchr(ip6addr, '/');
@@ -78,6 +103,19 @@ unsigned char  sipv6_prefixlength(const char *ip6addr)
     }
     return 0;
 }
+
+static bool is_hex(char c)
+{
+    // 'A' (0x41) and 'a' (0x61) are mapped in the ASCII table in such a way that masking the 0x20 bit turn 'a' in 'A'
+    if ((c & ~0x20) >= 'A' && (c & ~0x20) <= 'F')
+        return true;
+
+    if (c >= '0' && c <= '9')
+        return true;
+
+    return false;
+}
+
 static uint16_t hex(const char *p)
 {
     uint16_t val = 0;
--- a/test/libService/unittest/stoip6/stoip6test.cpp
+++ b/test/libService/unittest/stoip6/stoip6test.cpp
@@ -42,11 +42,10 @@ TEST(stoip6, TooShort)
 {
     char *addr = "FFFF:FFFF:";
     uint8_t ip[16];
-    uint8_t correct[16] = {0xff, 0xff, 0xff, 0xff, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };
-    // This should sto parsing when too short address given, the buffer hoewever is filled to that long
-    // So basically there is no error handling. We just check that first FFFF:FFFF gets filled and not trash after that.
-    // Rest should be filled with zeroes
-    stoip6(addr, strlen(addr), ip);
+    uint8_t correct[16] = {0};
+    // This should stop parsing when too short address given.
+    // Despite partial parsing, the entire buffer should be filled with zeroes
+    CHECK(false == stoip6(addr, strlen(addr), ip));
     CHECK(0 == memcmp(ip, correct, 16));
 }
 
@@ -58,7 +57,7 @@ TEST(stoip6, TooLongString)
     uint8_t correct[16] = {0};
     // This should not fill anything, too long string.
     // This is basically only validation we do
-    stoip6(addr, strlen(addr), ip);
+    CHECK(false == stoip6(addr, strlen(addr), ip));
     CHECK(0 == memcmp(ip, correct, 16));
 }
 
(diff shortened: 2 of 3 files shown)


source/libip6string/stoip6.c
```diff
@@ -66,10 +70,12 @@ void stoip6(const char *ip6addr, size_t len, void *dest)
         memmove(addr + head_size + inserted_size, addr + head_size, tail_size);
         memset(addr + head_size, 0, inserted_size);
     } else if (field_no != 8) {
-        /* Should really report an error if we didn't get 8 fields */
-        memset(addr, 0, 16 - field_no * 2);
+        //Report an error if we didn't get 8 fields
+        return false;
```
kjbracey-arm: Just in case, for backwards compatibility, keep the memset in. We've already written something to the buffer. May as well fully fill it. 

I'd like to minimise the chances of breaking a large body of code using this.

Although we now are being stricter on the `is_hex` check. Maybe that should carry on but set an "error" flag to return false at the end? Guarantee we don't break anyone?

source/libip6string/stoip6.c
```diff
@@ -37,15 +39,17 @@ void stoip6(const char *ip6addr, size_t len, void *dest)
...
     }
 
     // First go forward the string, until end, noting :: position if any
     for (field_no = 0, p = ip6addr; (len > (size_t)(p - ip6addr)) && *p && field_no < 8; p = q + 1) {
         q = p;
         // Seek for ':' or end
         while (*q && (*q != ':')) {
-            q++;
+            //There must only be hex characters besides ':'
```
kjbracey-arm: Space after `//`, also below.

source/libip6string/stoip6.c
```diff
@@ -20,15 +20,17 @@
 #include "ip6string.h"
 
 static uint16_t hex(const char *p);
+static bool is_hex(const char c);
```
kjbracey-arm: Not a fan of meaningless `const` in declaration prototypes - you're passing by value. You're free to retain it in the definition below, I guess, but there's no `const size_t` or `void * const dest` going on, so I think this is just randomly picked up from the line above where the `const` means something different. This is actually analogous to `const char * const p`.

(Argument variables can be declared const or not inside a function, indicating the constness of the copy. For the parameters, const is meaningless, and doesn't affect type-checking of definition versus declaration.).

source/libip6string/stoip6.c
Taiki-San: Ok with memset.

As for soft failing, I'm not totally confortable with such a lax parser, especially for something as important as parsing IP addresses. If you want to push to avoid breaking third-party code, I'll make that change but I'd rather see buggy/unsafe code break before being exploited by hackers. Your call.

kjbracey-arm: Tend to agree, but the current code is currently returning void, and there are lots of users, so even if you make it return false, no current code is going to be checking the return.

By returning early, you make current users process uninitialised data rather a definite value they're getting at the moment.

Once people actually start looking at the return value, then it becomes a strict parser.

Taiki-San: Would memset-ing the entire output buffer to 0 every time we return false be an appropriate alternative?

kjbracey-arm: Seems plausible. I'm just nervous about any change to the return values here, as I suspect there's not enough CI on this repo - we wouldn't see any problems until incorporated into upstream projects.

Hmm. Okay, make it clear when returning false. (Stylistically I'm fine with a `goto error` for that).

source/libip6string/stoip6.c
```diff
@@ -69,11 +68,15 @@ bool stoip6(const char *ip6addr, size_t len, void *dest)
...
-    } else if (field_no != 8) {
-        //Report an error if we didn't get 8 fields
-        return false;
+    } else if (field_no != 8) { // Report an error if we didn't get 8 fields
+        goto error;
     }
     return true;
+
+error:
+    // Fill the output buffer with 0 so we stick to the old failure mechanism
```
kjbracey-arm: I don't think this exactly match the old failure in all cases (eg non-hex digits), so probably best to not claim that.

Taiki-San: Is the new comment less ambiguous or should I rather drop the mention of the old behavior ?

source/libip6string/stoip6.c
```diff
@@ -43,18 +43,27 @@ bool stoip6(const char *ip6addr, size_t len, void *dest)
...
-    for (p = ip6addr; (len > (size_t)(p - ip6addr)) && *p && field_no < 8; p = q + 1) {
+    for (field_no = 0, p = ip6addr; (len > (size_t)(p - ip6addr)) && *p && field_no < 8; p = q + 1) {
         q = p;
         while (*q && (*q != ':')) { // Seek for ':' or end
             if (!is_hex(*q++)) { // There must only be hex characters besides ':'
                 goto error;
             }
         }
+
+        if((q - p) > 4) { // We can't have more than 4 hex digits per segments
```
kjbracey-arm: Formatting - space after if. "Per segment"

source/libip6string/stoip6.c
```diff
@@ -43,18 +43,27 @@ bool stoip6(const char *ip6addr, size_t len, void *dest)
...
+            goto error;
+        }
+
         // Convert and write this part, (high-endian AKA network byte order)
         addr = common_write_16_bit(hex(p), addr);
         field_no++;
+
         // Check if we reached "::"
         if ((len > (size_t)(q - ip6addr)) && *q && (q[0] == ':') && (q[1] == ':')) {
+            if(coloncolon != -1) { // We are not supposed to see "::" more than once per address
```
kjbracey-arm: Formatting - space after if

Taiki-San:
Rebased the patches after fixing the formatting you mentioned.
Also tweaked the unit tests: it's now checking that the call succeed/failed and a new tests validate the failure in a few scenarios.

Taiki-San:
I don't have the toolchain to run the tests locally, but are they ran by the CI? 504b8b2 should have failed.

kjbracey-arm:
I fear the CI is not running those unit tests on PRs. This repo is low-traffic enough we've not spent much time on automation. I can run them locally.

test/libService/unittest/stoip6/stoip6test.cpp
```diff
@@ -96,13 +94,31 @@ TEST(stoip6, RegressionTestForOffByOne)
...
+{
+    uint8_t ip[16];
+    uint8_t correct[16] = {0};
+
+    const char* invalidArray[] =
+    {
+        "FFFF:FFFF::FFFF::FFFF", // Two ::
+        "F:F:F:FqF:F:F:F:F",     // Non-hex character
+        "F:F:F:FFFFF:F:F:F:F"    // >4 hex characters in a segment
+    }
```
kjbracey-arm: Missing ;

test/libService/unittest/stoip6/stoip6test.cpp
```diff
@@ -96,13 +94,31 @@ TEST(stoip6, RegressionTestForOffByOne)
...
     free(sourceTemp);
 }
 
+// Test various illegal formats to ensure proper rejection
+TEST(stoip6, InvalidAddresses)
+{
+    uint8_t ip[16];
+    uint8_t correct[16] = {0};
+
+    const char* invalidArray[] =
```
kjbracey-arm: Formatting

test/libService/unittest/stoip6/stoip6test.cpp
```diff
@@ -96,13 +94,31 @@ TEST(stoip6, RegressionTestForOffByOne)
...
+    {
+        "FFFF:FFFF::FFFF::FFFF", // Two ::
+        "F:F:F:FqF:F:F:F:F",     // Non-hex character
+        "F:F:F:FFFFF:F:F:F:F"    // >4 hex characters in a segment
+    }
+
+    for(uint8_t i = 0; i < 3; ++i) {
```
kjbracey-arm: Formatting

test/libService/unittest/stoip6/stoip6test.cpp
```diff
@@ -96,13 +94,31 @@ TEST(stoip6, RegressionTestForOffByOne)
...
+        CHECK(false == stoip6(invalidArray[i], strlen(invalidArray[i], ip)));
```
kjbracey-arm: Misplaced `)`

test/libService/unittest/stoip6/stoip6test.cpp
```diff
@@ -96,13 +94,31 @@ TEST(stoip6, RegressionTestForOffByOne)
...
+        CHECK(0 == memcmp(ip, correct, 16));
```
kjbracey-arm: This assert failed, don't know which iteration.

test/libService/unittest/stoip6/stoip6test.cpp
```diff
@@ -58,20 +57,19 @@ TEST(stoip6, TooLongString)
...
-    uint8_t ip[17] = {0};
-    uint8_t correct[17] = { 0, 0xff, 0, 0xff, 0, 0xff, 0, 0xff, 0, 0xff, 0, 0xff, 0xff, 0xff, 0xff, 0xff, 0};
-    // Again.. there is not really any error handling (no return value)
-    // Just make sure that it does not overflow
-    stoip6(addr, strlen(addr), ip);
-    CHECK(0 == memcmp(ip, correct, 17)); // Note, we are checking 17, to make sure one byte after address in not touched.
+    uint8_t ip[16] = {0};
+    uint8_t correct[16] = {0};
+
+    CHECK(false == stoip6(addr, strlen(addr), ip));
```
kjbracey-arm: Assertion failed.

test/libService/unittest/stoip6/stoip6test.cpp
```diff
@@ -42,11 +42,10 @@ TEST(stoip6, TooShort)
...
     uint8_t ip[16];
-    uint8_t correct[16] = {0xff, 0xff, 0xff, 0xff, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };
-    // This should sto parsing when too short address given, the buffer hoewever is filled to that long
-    // So basically there is no error handling. We just check that first FFFF:FFFF gets filled and not trash after that.
-    // Rest should be filled with zeroes
-    stoip6(addr, strlen(addr), ip);
+    uint8_t correct[16] = {0};
+    // This should stop parsing when too short address given.
+    // Despite partial parsing, the entire buffer should be filled with zeroes
+    CHECK(false == stoip6(addr, strlen(addr), ip));
```
kjbracey-arm: Assertion failed.

Taiki-San:
This should fix the errors. I apologize for taking so much of your time on those trivial issues.

Taiki-San:
Do you want me to rebase before merging?

kjbracey-arm:
Yes, squash it together as much as is sensible to make it neat.

Taiki-San:
Just noticed we're fairly lax in regard to the `len` argument. Specifically, if the string isn't NULL terminated, we'll over-read it.
Not sure if we should be more aggressive with stopping/reporting error, as there is a high chance that this will break third-party code, thanks to off-by-one bugs.

kjbracey-arm:
Probably worth tightening that. The primary use case for specifying length is when we're parsing things like `http://[fe80::1]` or `fe80::/64` - some other layer has already identified a separator. So normally that next byte would be something that would stop the parse. I wouldn't expect someone to have a stale hex digit in the terminator position, but I guess we should be robust.

And hopefully everyone who is parsing stuff like the above is passing an accurate length showing the separator, or they'll now get 0.

source/libip6string/stoip6.c
```diff
@@ -64,12 +65,13 @@ bool stoip6(const char *ip6addr, size_t len, void *dest)
...
 
         // Check if we reached "::"
-        if ((len > (size_t)(q - ip6addr)) && *q && (q[0] == ':') && (q[1] == ':')) {
+        if (len && (q[0] == ':') && (q[1] == ':')) {
             if (coloncolon != -1) { // We are not supposed to see "::" more than once per address
                 goto error;
             }
             coloncolon = field_no;
             q++;
+            len -= 1;
```
kjbracey-arm: These should probably be `len--` just to match surrounding style.

Taiki-San:
Done

Taiki-San:
@kjbracey-arm Would you mind merging this PR? Once it's done, I'll be able to update https://github.com/ARMmbed/mbed-os/pull/6293 to use those new APIs.

kjbracey-arm:
Reverted due to some system test failures, including in unit tests. (Actual PR checks are minimal on this repo, so not spotted earlier). @artokin will give some details.

When he does, please add the failures to the unit tests, and make sure you run them.

artokin:
During the test it failed to convert 2001::0/64 [32:30:30:31:3a:3a:30:2f:36:34],  is_hex is failing with character 2f.
Another failure happens with address: [66:65:38:30:3a:3a:31]

When running unittest I get the following failure: 

```
./stoip6_unit_tests 
..................
stoip6test.cpp:37: error: Failure in TEST(stoip6, ZeroAddress)
	CHECK(0 == memcmp(ip, ipv6_test_values[i].bin, 16)) failed

.
Errors (1 failures, 19 tests, 19 ran, 42 checks, 0 ignored, 0 filtered out, 0 ms)

```

kjbracey-arm:
I did believe that first one should fail, because it was up to the caller to detect the `'/'` before-hand, eg via `sipv6_prefixlength`, if it's a prefix context, and exclude it by restricting `len`, rather than the conversion function quietly stopping on `'/'`.

Or

artokin:
Unit test passed now. I'm running other test to verify the functionality.

PR #74, opened 2018-09-26
Make stoip6 return whether the conversion succeed (fix a bug from #72)

Taiki-San:
This pull request fix a bug introduced in the refactor of #72.
We could drive `len` to 0 within the loop and decrement it once too many with the `for` loop.
This pull request fix the issue and add a new unit test case.

It appears unlikely the previous unit tests would have run with the old code, so no new radically different unit test appear to be necessary.

I don't have the infrastructure to run the tests on my machine, so could someone have a go?

### DIFF
--- a/source/libip6string/stoip6.c
+++ b/source/libip6string/stoip6.c
@@ -20,6 +20,7 @@
 #include "ip6string.h"
 
 static uint16_t hex(const char *p);
+static bool is_hex(char c);
 
 /**
  * Convert numeric IPv6 address string to a binary.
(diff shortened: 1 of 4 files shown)


kjbracey-arm:
Please rebase, making sure you adjust the new "prefix" function to return your new return value.

Taiki-San:
Can I get the detail of the failure?

artokin:
I'm sorry, this was my mistake. All needed pieces were not merged to our CI. They are merged now and CI job is re-triggered.

Taiki-San:
@kjbracey-arm do you think this PR need anything else?

source/libip6string/stoip6.c
```diff
@@ -101,17 +131,32 @@ int stoip6_prefix(const char *ip6addr, void *dest, int_fast16_t *prefix_len_out)
...
             }
 
             *prefix_len_out = prefix_length;
         }
     }
 
-   stoip6(ip6addr, addr_len, dest);
+    if (!stoip6(ip6addr, addr_len, dest)) {
+        /* parser failure */
+        return -2;
```
kjbracey-arm: Don't think it's worth distinguishing the precise malformedness - would suggest returning -1 like the other cases above. 

If you really want it distinguished, you would need to document it to be useful.

Taiki-San: I don't really care. Thought the failure was different enough to warrant a different return code but this likely isn't very actionnable anyway.

PR #75, opened 2018-09-27
Add new function stoip6_prefix

artokin:
Added new method for parsing numeric IPv6 with prefix information to
binary format.

### DIFF
--- a/mbed-client-libservice/ip6string.h
+++ b/mbed-client-libservice/ip6string.h
@@ -76,7 +76,7 @@ unsigned char sipv6_prefixlength(const char *ip6addr);
  *
  * \param ip6addr IPv6 address in string format.
  * \param dest buffer for address. MUST be 16 bytes.
- * \param prefix_len_out length of prefix
+ * \param prefix_len_out length of prefix, is set to -1 if no prefix given
  *
  * \return 0 on success, negative value otherwise. prefix_len_out contains prefix length.
   */
--- a/mbed-client-libservice/ip6string.h
+++ b/mbed-client-libservice/ip6string.h
@@ -69,6 +69,19 @@ void stoip6(const char *ip6addr, size_t len, void *dest);
  */
 unsigned char sipv6_prefixlength(const char *ip6addr);
 
+/**
+ * Convert numeric IPv6 address string with prefix to a binary.
+ *
+ * IPv4 tunneling addresses are not covered.
+ *
+ * \param ip6addr IPv6 address in string format.
+ * \param dest buffer for address. MUST be 16 bytes.
+ * \param prefix_len_out length of prefix
+ *
+ * \return 0 on success, negative value otherwise. prefix_len_out contains prefix length.
+  */
+int stoip6_prefix(const char *ip6addr, void *dest, int16_t *prefix_len_out);
+
 #ifdef __cplusplus
 }
 #endif
--- a/mbed-client-libservice/ip6string.h
+++ b/mbed-client-libservice/ip6string.h
@@ -80,7 +80,7 @@ unsigned char sipv6_prefixlength(const char *ip6addr);
  *
  * \return 0 on success, negative value otherwise. prefix_len_out contains prefix length.
   */
-int stoip6_prefix(const char *ip6addr, void *dest, int16_t *prefix_len_out);
+int stoip6_prefix(const char *ip6addr, void *dest, int_fast16_t *prefix_len_out);
 
 #ifdef __cplusplus
 }
--- a/source/libip6string/stoip6.c
+++ b/source/libip6string/stoip6.c
@@ -80,23 +80,34 @@ unsigned char sipv6_prefixlength(const char *ip6addr)
     return 0;
 }
 
-int stoip6_prefix(const char *ip6addr, void *dest, int16_t *prefix_len_out)
+int stoip6_prefix(const char *ip6addr, void *dest, int_fast16_t *prefix_len_out)
 {
-    size_t addr_len;
+    size_t addr_len, total_len;
+    int_fast16_t prefix_length;
 
     if (prefix_len_out) {
         *prefix_len_out = -1;
     }
 
-    addr_len = strlen(ip6addr);
-    char *ptr = strchr(ip6addr, '/');
+    total_len = addr_len = strlen(ip6addr);
+    const char *ptr = strchr(ip6addr, '/');
     if (ptr) {
         addr_len = ptr - ip6addr;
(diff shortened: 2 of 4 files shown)


mbed-client-libservice/ip6string.h
```diff
@@ -69,6 +69,19 @@ void stoip6(const char *ip6addr, size_t len, void *dest);
...
 unsigned char sipv6_prefixlength(const char *ip6addr);
 
+/**
+ * Convert numeric IPv6 address string with prefix to a binary.
+ *
+ * IPv4 tunneling addresses are not covered.
+ *
+ * \param ip6addr IPv6 address in string format.
+ * \param dest buffer for address. MUST be 16 bytes.
+ * \param prefix_len_out length of prefix
```
kjbracey-arm: Document that this is filled as -1 if no prefix length given.

I note you permit this to be NULL. If it is NULL, is it best to just process and ignore the prefix length? Or could stoip6_prefix(addr, dest, NULL) actually be "just process address", so it faults if there is a prefix, and saves the external strlen?

source/libip6string/stoip6.c
```diff
@@ -78,6 +78,37 @@ unsigned char  sipv6_prefixlength(const char *ip6addr)
...
+    }
+
+   addr_len = strlen(ip6addr);
+   prefix_length = sipv6_prefixlength(ip6addr);
+   if (prefix_length) {
+       char *ptr = strchr(ip6addr, '/');
+       if (ptr) {
+           addr_len = ptr - ip6addr;
+           if (prefix_len_out) {
+               *prefix_len_out = (int16_t)prefix_length;
```
kjbracey-arm: Don't need this cast - unsigned char fits fully inside int16_t.

source/libip6string/stoip6.c
```diff
@@ -78,6 +78,37 @@ unsigned char  sipv6_prefixlength(const char *ip6addr)
     }
     return 0;
 }
+
+int stoip6_prefix(const char *ip6addr, void *dest, int16_t *prefix_len_out)
+{
+    size_t addr_len;
+    unsigned char prefix_length;
+
+    if (!ip6addr) {
```
kjbracey-arm: No need for null checks; rest of file doesn't do this.

test/libService/unittest/stoip6/stoip6test.cpp
```diff
@@ -221,3 +221,55 @@ TEST(stoip6_2, test_2_12)
...
 
+/***********************************************************/
+/* Third test group for stoip6_prefix */
+
+const char string_prefix_addr[][40] =
+{
+    "2001:db8::1:0:0:1/64",     // 1
+    "2001::/60",                // 2
+    "::1/48",                   // 3
+    "::/00",                    // 4
```
kjbracey-arm: See you're getting -1 here - that shouldn't happen, should be 0. Looks like that prefix_length function is annoying/broken, returning 0 if no prefix.

source/libip6string/stoip6.c
```diff
@@ -78,6 +78,37 @@ unsigned char  sipv6_prefixlength(const char *ip6addr)
...
+        return -1;
+    }
+
+    if (prefix_len_out) {
+        *prefix_len_out = -1;
+    }
+
+   addr_len = strlen(ip6addr);
+   prefix_length = sipv6_prefixlength(ip6addr);
+   if (prefix_length) {
```
kjbracey-arm: Hm, looks like sipv6_prefixlength is being bad. Need to distinguish "no prefix length" from "/0". Maybe pull in equivalent code inline, rather than call it? Would save strchr twice.

source/libip6string/stoip6.c
```diff
@@ -78,6 +78,37 @@ unsigned char  sipv6_prefixlength(const char *ip6addr)
...
+       char *ptr = strchr(ip6addr, '/');
+       if (ptr) {
+           addr_len = ptr - ip6addr;
+           if (prefix_len_out) {
+               *prefix_len_out = (int16_t)prefix_length;
+           }
+       }
+   }
+
+   stoip6(ip6addr, addr_len, dest);
```
kjbracey-arm: Must remember to update this to give return value later.

mbed-client-libservice/ip6string.h
artokin: Assuming that processing and ignoring the prefix_length is more flexible for the application.

source/libip6string/stoip6.c
```diff
@@ -70,14 +70,39 @@ void stoip6(const char *ip6addr, size_t len, void *dest)
...
+    if (prefix_len_out) {
+        *prefix_len_out = -1;
+    }
+
+    addr_len = strlen(ip6addr);
+    char *ptr = strchr(ip6addr, '/');
+    if (ptr) {
+        addr_len = ptr - ip6addr;
+        if (prefix_len_out) {
+            *prefix_len_out = strtoul(ptr + 1, 0, 10);
```
kjbracey-arm: Maybe be a little bit stricter here, in the way that prefixlength isn't? Check that we consumed the entire remaining string, and it's in the range 0-128?

source/libip6string/stoip6.c
```diff
@@ -70,14 +70,39 @@ void stoip6(const char *ip6addr, size_t len, void *dest)
...
+int stoip6_prefix(const char *ip6addr, void *dest, int16_t *prefix_len_out)
+{
+    size_t addr_len;
+
+    if (prefix_len_out) {
+        *prefix_len_out = -1;
+    }
+
+    addr_len = strlen(ip6addr);
+    char *ptr = strchr(ip6addr, '/');
```
kjbracey-arm: const char *. (This is one of the const holes in C - C++ fixes it with an overload).

mbed-client-libservice/ip6string.h
```diff
@@ -69,6 +69,19 @@ void stoip6(const char *ip6addr, size_t len, void *dest);
...
+ *
+ * IPv4 tunneling addresses are not covered.
+ *
+ * \param ip6addr IPv6 address in string format.
+ * \param dest buffer for address. MUST be 16 bytes.
+ * \param prefix_len_out length of prefix, is set to -1 if no prefix given
+ *
+ * \return 0 on success, negative value otherwise. prefix_len_out contains prefix length.
+  */
+int stoip6_prefix(const char *ip6addr, void *dest, int16_t *prefix_len_out);
```
kjbracey-arm: Maybe int_fast16_t, just because we're using uint_fast8_t above?

PR #76, opened 2018-10-28
Compiler warning fixes

TeroJaasko:
These two commits silence a bunch of GCC warnings on a non-standard environment.

### DIFF
--- a/mbed-client-libservice/ns_types.h
+++ b/mbed-client-libservice/ns_types.h
@@ -120,7 +120,7 @@ typedef int_fast32_t int_fast24_t;
 #if defined __CC_ARM || defined __TASKING__
 #define alignas(n) __align(n)
 #define __alignas_is_defined 1
-#elif (__STDC_VERSION__ >= 201112L) || (defined __cplusplus && __cplusplus >= 201103L)
+#elif (defined __STDC_VERSION__ && __STDC_VERSION__ >= 201112L) || (defined __cplusplus && __cplusplus >= 201103L)
 #include <stdalign.h>
 #elif defined __GNUC__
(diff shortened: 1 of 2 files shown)


mbed-client-libservice/ns_types.h
```diff
@@ -120,7 +120,7 @@ typedef int_fast32_t int_fast24_t;
 #if defined __CC_ARM || defined __TASKING__
 #define alignas(n) __align(n)
 #define __alignas_is_defined 1
-#elif (__STDC_VERSION__ >= 201112L) || (defined __cplusplus && __cplusplus >= 201103L)
+#elif (defined __STDC_VERSION__ && __STDC_VERSION__ >= 201112L) || (defined __cplusplus && __cplusplus >= 201103L)
```
kjbracey-arm: Ew. Glad this shuts up the warning though - I've worked on compilers with complain anyway as the substitution happens before seeing that the `&&` expression short-circuits evaluation.

As it happens, this is known to fail on ARMC5 in "C++11" mode - it does indicate C++11 version without providing `<stdalign.h>`. Sigh. It's the best bit.

TeroJaasko: stdalign.h was missing on our environment also and I had to add it to get even this far. 
Would it be possible to have some kind of manual override there, eg. "#define HAVE_STDALIGN_H 0" or "#define FORCE_MANUAL_ALIGN 1" would make the ns_types.h to define the alignas() manually, without external headers?

TeroJaasko:
Could you merge this or do you mind if I do it?

TeroJaasko:
Kärpät kiittää. :)

PR #78, opened 2018-11-12
Avoid memcmp(x, NULL, 0)

kjbracey-arm:
libBits' comparison and copying operations are assumed to do the right
thing when passed a NULL pointer and zero length by various bits of
code.

It is not legal to call any standard C library function like memcmp or
memcpy with an invalid pointer, even if length is 0.

Copy code avoids that, but bitsequal could call memcmp(x, NULL, 0). Add
an extra condition to avoid this.

Spotted while inspecting Coverity reports (although it wasn't directly
related to the memcmp call, but a false positive on another path).


kjbracey-arm:
@artokin

PR #80, opened 2019-01-15
Added possibility for monitor heap free sector count and read sectors.

juhhei01:


juhhei01:
@kjbracey-arm , @JarkkoPaso , @AnttiKauppila Please review this memory fragmentation monitor

juhhei01:
Now we accept 2 sizeof(int) memory holes. Sounds a very small should update this by little a bit bigger one. Update is mandatory if we start see very small entries at free list

PR #84, opened 2019-02-04
Support C++03 compilers that cannot return properly-typed pointers.

SeppoTakalo:
IAR 7 seems to be one of these compilers.


SeppoTakalo:
Equivalent change already proposed to Mbed OS in https://github.com/ARMmbed/mbed-os/pull/9424

SeppoTakalo:
Had to re-apply the patch, as there was 3 lines missing.. now this is equivalent of what is already applied in Mbed OS.

PR #85, opened 2019-02-18
Sync with Mbed OS

artokin:
Fetch latest changes from Mbed OS


artokin:
@mikter , @deepakvenugopal would you please review?

PR #86, opened 2019-06-07
nsdynmenlib API to add memory region to heap

artokin:
Add new API to nsdynmemlib that allows user to add memory region to
already initialized heap.

Two additional memory regions can be added to nsdynmemlib. Amount of
regions can be adjusted with internal macro REGION_COUNT.


artokin:
@mikter , @kjbracey-arm , @SeppoTakalo would you please review?

PR #93, opened 2020-11-11
Added dynamic memory tracker and hooks to dynmem library

mikaleppanen:
NS dynamic memory tracker tracks memory allocations. It has following
interface:

ns_dyn_mem_tracker_lib_alloc: is called on alloc

ns_dyn_mem_tracker_lib_free: is called on free

ns_dyn_mem_tracker_lib_step: updates the lifetime of individual memory
blocks, and it used to detected which blocks are permanent and which
have been allocated lately

ns_dyn_mem_tracker_lib_allocator_lists_update: updates allocator lists,
there are lists for top allocators, permanent allocators and allocators
going to permanent allocators list

ns_dyn_mem_tracker_lib_max_snap_shot_update: can be used to update memory
maximum usage snap shot, this can be called on alloc if snap shot is
needed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ source/libip6string/ip6tos.c \
 source/libip6string/stoip6.c \
 source/libList/ns_list.c \
 source/nsdynmemLIB/nsdynmemLIB.c \
+source/nsdynmemtracker/nsdynmem_tracker_lib.c \
 source/nvmHelper/ns_nvm_helper.c \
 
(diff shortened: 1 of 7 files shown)


mikaleppanen:
I'll still make optimizations at least to list for loops etc. but main structure / interface should be here.

mikter (approved):

juhhei01 (approved):
Looks good

PR #96, opened 2020-12-08
Enable heap stats available always.

juhhei01:
Heap stats request works now with any any user. Data read is working without init enable stats

### DIFF
--- a/mbed-client-libservice/nsdynmemLIB.h
+++ b/mbed-client-libservice/nsdynmemLIB.h
@@ -67,7 +67,6 @@ typedef struct mem_stat_t {
     uint32_t heap_alloc_fail_cnt;               /**< Counter for Heap allocation fail. */
 } mem_stat_t;
 
-
 typedef struct ns_mem_book ns_mem_book_t;
 
 /**
--- a/source/nsdynmemLIB/nsdynmemLIB.c
+++ b/source/nsdynmemLIB/nsdynmemLIB.c
@@ -48,6 +48,7 @@ struct ns_mem_book {
     ns_mem_heap_size_t temporary_alloc_heap_limit;   /* Amount of reserved heap temporary alloc can't exceed */
 };
 
+static mem_stat_t default_stats;
 static ns_mem_book_t *default_book; // heap pointer for original "ns_" API use
 
 // size of a hole_t in our word units
@@ -163,12 +164,14 @@ ns_mem_book_t *ns_mem_init(void *heap, ns_mem_heap_size_t h_size,
     ns_list_init(&book->holes_list);
     ns_list_add_to_start(&book->holes_list, hole_from_block_start(book->heap_main[0]));
 
-    book->mem_stat_info_ptr = info_ptr;
-    //RESET Memory by Hea Len
     if (info_ptr) {
-        memset(book->mem_stat_info_ptr, 0, sizeof(mem_stat_t));
-        book->mem_stat_info_ptr->heap_sector_size = book->heap_size;
+        book->mem_stat_info_ptr = info_ptr;
+    } else {
+        book->mem_stat_info_ptr = &default_stats;
     }
+    //RESET Memory by Heap Len
+    memset(book->mem_stat_info_ptr, 0, sizeof(mem_stat_t));
+    book->mem_stat_info_ptr->heap_sector_size = book->heap_size;
     book->temporary_alloc_heap_limit = book->heap_size / 100 * (100 - TEMPORARY_ALLOC_FREE_HEAP_THRESHOLD);
 #endif
     //There really is no support to standard malloc in this library anymore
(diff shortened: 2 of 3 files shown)


juhhei01:
@mikter , @mikaleppanen , @JarkkoPaso , @artokin and @kjbracey-arm Please review this generic heap stat update.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -124,6 +126,25 @@ const mem_stat_t *ns_dyn_mem_get_mem_stat(void)
...
 
+int ns_dyn_mem_read_mem_stat(mem_gen_stat_t *mem_gen_stat)
+{
+#ifndef STANDARD_MALLOC
+    if (!default_book) {
+        return -1;
+    }
+
+    platform_enter_critical();
+    mem_gen_stat->heap_sector_size = default_book->heap_size;
```
mikter, line 137: Does this count the memory off all separate books together?

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -442,12 +465,18 @@ static void *ns_mem_internal_alloc(ns_mem_book_t *book, const ns_mem_block_size_
...
 done:
-    if (book->mem_stat_info_ptr) {
-        if (block_ptr) {
-            //Update Allocate OK
+
+
+    if (block_ptr) {
+        //Update Allocate OK
+        book->heap_alloc_total_bytes += (data_size + 2) * sizeof(ns_mem_word_size_t);
+        if (book->mem_stat_info_ptr) {
```
mikter, line 473: How about having the null check in the dev_stat_update function to reduce code size

source/nsdynmemLIB/nsdynmemLIB.c
juhhei01: when you add more Sectors they are added to this size.

mbed-client-libservice/nsdynmemLIB.h
```diff
@@ -67,6 +67,16 @@ typedef struct mem_stat_t {
     uint32_t heap_alloc_fail_cnt;               /**< Counter for Heap allocation fail. */
 } mem_stat_t;
 
+/**
+ * /struct mem_gen_stat_t
```
mikter, line 71: Is this just extra complexity why was the previous stat enough? it does not look like you are saving much

source/nsdynmemLIB/nsdynmemLIB.c
juhhei01: Good point is it eating performance then. It is some kind of balancing.

mbed-client-libservice/nsdynmemLIB.h
juhhei01: You forget that user must enable those previous stats. This change enable possibility to get minimal information always at any heap user. Old system works when app enable heap.

juhhei01: and that old stat structure is giving by user.

mikter: I did not forget that was the point why not always enable it. because this just brings complexity without saving anything

mikter: So my suggestion is that by default always enable stats and if user gives it then use the structure from the user and not the default internal

juhhei01: That is then API change. Now user reset some of the values at structure I believe.  Actually simulator port use own structure if user not gives so it works then Like you want. const pointer can't edit any values. But then it actually works like we want. I start like your idea. Lets make that like at simulator platform.

source/nsdynmemLIB/nsdynmemLIB.c
```diff
@@ -124,6 +126,25 @@ const mem_stat_t *ns_dyn_mem_get_mem_stat(void)
...
+{
+#ifndef STANDARD_MALLOC
+    if (!default_book) {
+        return -1;
+    }
+
+    platform_enter_critical();
+    mem_gen_stat->heap_sector_size = default_book->heap_size;
+    mem_gen_stat->heap_alloc_fail_cnt = default_book->heap_alloc_fail_cnt;
+    mem_gen_stat->heap_alloc_total_bytes = default_book->heap_alloc_total_bytes;
```
artokin, line 139: NULL check is missing from parameter mem_gen_stat.

The signature of this API is:
` int ns_dyn_mem_read_mem_stat(mem_gen_stat_t *mem_gen_stat)`
And the other related API is:
` const mem_stat_t *ns_dyn_mem_get_mem_stat(void)`
Should the API naming be aligned? 

Now we are having two methods for the same purpose. Why we need two?

juhhei01:
@artokin and @miklter Please review. Stats are now available always. without any API changes

mikaleppanen (approved):

artokin (commented):
Looks much better now!
Should ns_dyn_mem_get_mem_stat description be updated?

mikter (approved):

